PR #4, opened 2015-05-18
disable norms on _all field as per

Mpdreamz:
updated logstash template https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/91


mivano:
Added this change manually (https://github.com/serilog/serilog-sinks-elasticsearch/commit/ff16150fa057927889c4f2aac535bc99d98e4fb2) and updated tests.

PR #5, opened 2015-05-18
Fix for issue #451 (ES crash)

tnicolaysen:
The new code basically does two things;

1) Wraps the entire template creating in a try/catch where the exception is written to the self log.
2) It then keeps state if the template was created. It will attempt to create the template until it has managed to do so. 

This last part needs a sanity check, but I think self-recovery is a good strategy as it's the only way to get it working again without restarting the app(/pool).

The new code/behavior has some tests, but I didn't find a way to test the recovery functionality.

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.ElasticSearch.nuspec
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.ElasticSearch.nuspec
@@ -8,7 +8,7 @@
     <language>en-US</language>
     <projectUrl>http://serilog.net</projectUrl>
     <licenseUrl>http://www.apache.org/licenses/LICENSE-2.0</licenseUrl>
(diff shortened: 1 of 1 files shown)


mivano:
Sorry, was away on holiday and did not have access to a PC. Thanks for the PR, looks good. Regarding 2; if the system is unable to add the template before it starts writing the first log events and for some reason the events are written to the index, the mapping is done implicitly. Changing it afterwards might fail (data types might not match) and as such the registration will fail again. And we ran in a loop. Maybe better to do only once? 

@Mpdreamz You have a suggestion about this?

btw can you pull against the dev branch instead?

tnicolaysen:
I'm going to submit a new PR on the right branch, and I'm also going to leave out this new behavior and submit that as a new issue (that can be tagged with enhancement).

I've created a new issue for this and we can continue the discussion there.
https://github.com/serilog/serilog/issues/453

PR #8, opened 2015-06-15
Uses the IndexDecider to decide index name when pushing to index.

jarlelin:
Fixes an issue where the index decider was not properly used with pusing events thorugh the ElasticLogShipper.


mivano:
Thanks for the fix! Package should be available soon.

PR #9, opened 2015-06-25
Added overload for appSettings configuration to resolve #425

matthewrwilton:
Added a simple overload with the parameters @mivano suggested. I made the index format and template registration optional to remain in the spirit of bare minimum configuration (server address only) required.

If this gets merged in the "Breaking changes for version 2" section in the readme should probably be updated to reflect that basic appSettings configuration is possible.

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
+++ b/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
@@ -17,7 +17,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Serilog.Configuration;
 using Serilog.Core;
-using Serilog.Debugging;
 using Serilog.Events;
 using Serilog.Sinks.Elasticsearch;
 
@@ -75,17 +74,8 @@ namespace Serilog
             string indexFormat = null,
             string templateName = null)
         {
-            ElasticsearchSinkOptions options;
-            try
-            {
-                IEnumerable<Uri> nodes = nodeUris.Split(',').Select(uriString => new Uri(uriString));
-                options = new ElasticsearchSinkOptions(nodes);
-            }
-            catch (Exception ex)
-            {
-                SelfLog.WriteLine("Exception while parsing node URIs from {0}: {1}", nodeUris, ex);
-                options = new ElasticsearchSinkOptions(new[] { new Uri(DefaultNodeUri) });
-            }
+            IEnumerable<Uri> nodes = nodeUris.Split(',').Select(uriString => new Uri(uriString));
+            var options = new ElasticsearchSinkOptions(nodes);
 
             if (!string.IsNullOrWhiteSpace(indexFormat))
             {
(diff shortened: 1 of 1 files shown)


src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
```diff
@@ -56,5 +56,35 @@ public static class LoggerConfigurationElasticsearchExtensions
...
+        /// Overload to allow basic configuration through AppSettings.
+        /// </summary>
+        /// <param name="loggerSinkConfiguration">Options for the sink.</param>
+        /// <param name="uri">A URI for the Elasticsearch instance</param>
+        /// <param name="indexFormat"><see cref="ElasticsearchSinkOptions.IndexFormat"/></param>
+        /// <param name="templateName"><see cref="ElasticsearchSinkOptions.TemplateName"/></param>
+        /// <returns>LoggerConfiguration object</returns>
+        public static LoggerConfiguration Elasticsearch(
+            this LoggerSinkConfiguration loggerSinkConfiguration,
+            string uri,
```
mivano: We can also pass in a Uri directly. This is supported by the AppSettings config (https://github.com/serilog/serilog/blob/dev/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs#L150).

src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
```diff
@@ -56,5 +56,35 @@ public static class LoggerConfigurationElasticsearchExtensions
...
+            string indexFormat = null,
+            string templateName = null)
+        {
+            var options = new ElasticsearchSinkOptions(new[] { new Uri(uri) });
```
mivano: Although if we use a string, we might be able to also allow passing in a collection or urls. So do a split on ; or , and pass those in as Uri array.

mivano:
Nice addition. See my inline comments. Not a big deal, but might be worthwhile to consider as changes afterwards are more difficult.

matthewrwilton:
I did wonder about adding support for multiple nodes and since you've brought it up it seems like it is worth adding. I'll add that in and resubmit.

matthewrwilton:
@mivano I added support for configuring multiple nodes using a comma separated list. I also realised that an exception could be thrown if the URI is invalid and I added handling to suppress this (which is how I think Serilog sinks are expected to behave).

mivano:
Thanks. I m just wondering if we should accept an incorrect Uri and silently use the default value when there is an exception. Normally the registration of a sink happens at startup and the other sinks throw an exception (see e.g. https://github.com/serilog/serilog/blob/dev/src/Serilog.FullNetFx/Sinks/RollingFile/RollingFileSink.cs#L64). So a misconfiguration can now lead to a localhost connection. Which might not be there and consequential logging will fail. The selflog is not very discoverable. 

@nblumhardt you have an opinion about this?

Besides this, a nice addition as most people only specify a (collection of) uri(s).

nblumhardt:
I'd favour consistency with the other sinks in this case - keeps things predictable.

matthewrwilton:
No worries. Updated to throw exceptions for invalid URIs.

mivano:
I merged this and a new package will be out soon. Documentation has been updated too. Thanks for the PR!

matthewrwilton:
Happy to help out :smile:

PR #12, opened 2015-08-12
delay auto-register-template-if-necessary call to the first timer tick

mookid8000:
when using the durable sink, because we should not do this synchronously on the initializing thread when logging is initialized


mivano:
Thanks looks good. Will merge to master.

PR #13, opened 2015-08-13
Update README.md to include an appSettings example

sstorie:
I needed to look at the code to understand how to use the appSettings reader with this sink, so I wanted to provide a little documentation to help others out.


mivano:
Great addition. thanks!

PR #14, opened 2015-08-20
Comment unused GetValue on WatsonBuckets - Fix Mono SerializationExce…

klettier:
The call to get WatsonBuckets throw a SerializationException on Mono (not found).
So the event is never send to EL.


mivano:
Thanks for the report. As it is an unused variable anyway it makes sense not to query for it at all.

PR #15, opened 2015-09-15
Do not depend on specific version of serilog

tilign:
Without this change this sink does not work for serilog 1.5.* w/o a redirect


Mpdreamz:
Thanks for the PR!

@nblumhardt has final say on this IMO. Most of the other sinks (atleast all I checked) do not set specificversion either.

nblumhardt:
Looks fine to me, though I haven't had any luck avoiding binding redirects this way - as I understand it, this setting only applies when the ES sink itself is compiled.

merbla:
Not sure if this file for the build is in use however might be worth bumping to `1.5` as a minimum.

https://github.com/serilog/serilog-sinks-elasticsearch/blob/master/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.ElasticSearch.nuspec#L14

PR #17, opened 2015-11-07
Added a way to derive from ElasticsearchSink class so that customized…

konste:
… EmitBatch method would be able to analyse Elasticsearch response and get specific errors in case emit was unsuccessful.

Unfortunately existing call _state.Client.Bulk(payload); does not allow to check Elasticsearch response in any way. I have split EmitBatch to two methods:
EmitBatchChecked does what EmitBatch did, but returns ES response.
Existing EmitBatch just calls EmitBatchChecked and drops return value.

Class which overrides ElasticsearchSink may override EmitBatch to call base class EmitBatchChecked, but then do necessary ES response analysis.

Konstantin

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.ElasticSearch.Symbols.nuspec
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.ElasticSearch.Symbols.nuspec
@@ -1,0 +1,23 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>Serilog.Sinks.Elasticsearch</id>
+    <version>2.0.60.2</version>
+    <authors>Michiel van Oudheusden, Martijn Laarman, Mogens Heller Grabe</authors>
+    <description>The perfect way for .NET apps to write structured log events to Elasticsearch.</description>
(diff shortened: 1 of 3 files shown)


mivano:
A long time item on the wishlist; what to do with errors returned from ES. Sounds like a good way to at least being able to see them if you derive. Would be nicer if we had a solution out of the box, but still no real good clue how. This looks like a non breaking change but it would be nice if others also had a look. So I wait for feedback before I merge to master.

I requeued the build and it appears to work now.

konste:
I have the code which parses ES response and extract errors from there. The question is how to handle that information so it is suitable for most users of ES sink. For my case I have created secondary logger (which sends log events to Seq) and when there are ES errors I send them to that secondary logger. I can add this code to pull request if it makes sense, but I'm afraid to make the whole thing more controversial. May be I better do it as a separate pull request.

mivano:
Another option might be to add an event/function etc to the sink that gets called when there are errors from the elasticsearch server. Then the calling code can attach to that and handle it. But I also see your approach. As there are no other comments till now, I will merge it.

konste:
Server response may be of interest even when there are no errors. I'd rather not limit usage options to errors only case. On the other hand the code to parse server response to extract specific errors (if any) appeared to be not trivial, which IMHO is caused by using low level client instead of Nest. So I would provide helper public function to parse server response. I have one and after cleaning it up will send another pull request for it.

PR #20, opened 2015-12-27
Rebuild Serilog.Sinks.Elasticsearch with the latest version of depend…

konste:
…encies to avoid the need for Serilog binding redirects in each project consuming this sink.

This change should help numerous build issues we have with Serilog.Sink.Elasticsearch and serve as a foundation for coming improvements in ES response handling.

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.ElasticSearch.nuspec
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.ElasticSearch.nuspec
@@ -11,12 +11,12 @@
     <iconUrl>http://serilog.net/images/serilog-sink-nuget.png</iconUrl>
     <tags>serilog logging elasticsearch</tags>
     <dependencies>
-      <dependency id="Serilog" version="[1.4.204,2)" />
-      <dependency id="Elasticsearch.Net" version="1.1.2" />
+      <dependency id="Serilog" version="[1.5.14,2)" />
+      <dependency id="Elasticsearch.Net" version="1.7.1" />
     </dependencies>
   </metadata>
   <files>
-    <file src="bin40\Release\Serilog.Sinks.Elasticsearch.dll" target="lib\net40" />
-    <file src="bin40\Release\Serilog.Sinks.Elasticsearch.xml" target="lib\net40" />
+    <file src="bin\Release\Serilog.Sinks.Elasticsearch.dll" target="lib\net45" />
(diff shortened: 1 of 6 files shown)


src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.ElasticSearch.nuspec
```diff
@@ -11,12 +11,12 @@
...
-      <dependency id="Serilog" version="[1.4.204,2)" />
-      <dependency id="Elasticsearch.Net" version="1.1.2" />
+      <dependency id="Serilog" version="[1.5.14,2)" />
+      <dependency id="Elasticsearch.Net" version="1.7.1" />
     </dependencies>
   </metadata>
   <files>
-    <file src="bin40\Release\Serilog.Sinks.Elasticsearch.dll" target="lib\net40" />
-    <file src="bin40\Release\Serilog.Sinks.Elasticsearch.xml" target="lib\net40" />
+    <file src="bin\Release\Serilog.Sinks.Elasticsearch.dll" target="lib\net45" />
```
nblumhardt: Dropping support for .NET 4? (The 4.5 binaries are included in the package by default.)

konste: I was not sure how exactly support for net40 is done. Are you saying those two lines should stay as they were? My concern is that the sink's own packages.config (there is only one) always refers to net45 version of Serilog and SerilogFullNetFx, so it seems like what this sink builds as net40 still requires 4.5! 

In short - it seems to me that Serilog.Sinks.Elasticsearch already dropped net40 silently. I may be wrong though.

Do you know what specific nuget command is used to build the package with both binaries? I don't see it among the sources.

konste: I just discovered Build.ps1. Taking another look.

konste: Ok, those two lines are reverted. Thanks for catching that quickly!

nblumhardt:
Thanks @konste - looks like CI is failing however?

konste:
Wow, I finally pushed it through!

nblumhardt:
LGTM - @mivano ?

mivano:
Looks fine to me. :+1:

konste:
Great! Pull it, please! I'm so tired of keeping those pesky binding redirects from Serilog 1.4 to 1.5 in about 50 projects only for that sink.

PR #21, opened 2016-02-02
Added dnx core version

alexvaluyskiy:
Todo:
- convert tests to xunit


ravengerUA:
I've almost done with this PR. I just need to do these things
- configure AppVeyor (disable tests auto discovery)
- find out how to deserialize exceptions in NET Core
- fix test "ShouldRegisterTheCorrectTemplateOnRegistration" on AppVeyor

nblumhardt:
If you provide an appveyor.yml file it needs to be complete (it overrides the entire build process). Here's the current one (improvements needed :-) )

```yml
version: '{build}'
build_script:
- ps: ./Build.ps1 -majorMinor "2.0" -patch "$env:APPVEYOR_BUILD_VERSION" -customLogger "C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
artifacts:
- path: Serilog.*.nupkg
deploy:
- provider: NuGet
  api_key:
    secure: nvZ/z+pMS91b3kG4DgfES5AcmwwGoBYQxr9kp4XiJHj25SAlgdIxFx++1N0lFH2x
  on:
    branch: master
```

Mpdreamz:
ty @ravengerUA :+1: I'm out for the next 9 days so I wont have chances to review. Also thinking we should merge this in a 3.0 branch while serilog 2.0 bakes. @mivano  thoughts?

mivano:
Nice! Would be great having this sink as a dnx version. I have no clue how to package this actually. Would this be a new package, packaged inside the existing package or a bump of the version?

mivano:
Since we merged some other PRs and made the sink compatible with ES 2 and removed the .NET 4 libs, this now has a merge conflict. @alexvaluyskiy willing to pick it up and see if we can provide a dnx version too?

gdoron:
Hi @mivano,

Is there an estimation when it will be possible to use the elasticsearch sink with dnx?
I don't even need dnxcore at the moment, full dotnet framework will do for now.
I got 14 days of free elasticsearch, but I'm not able to use it because of the lack of a working sink :cry: 

(Sorry for not knowing enough about ASP.NET core and not being able to assist here)

Thanks!

mivano:
Sorry, I have no estimate on this. My experience with dnx is minima. It would certainly be nice to have a combined dnx version in the package, but no progress so far except for this PR. @alexvaluyskiy any progress or update on this?

asrikantan80:
Hi @mivano @alexvaluyskiy Any idea when elastic search sink for serilog would be available for DNX?

nblumhardt:
Hi folks; Serilog 2 should ship in not much more than a month - I think we're down to the wire as far as getting a compatible ES sink out with it, so it's probably time to move on, and/or have another attempt. ( @alexvaluyskiy please do jump in if you have anything in the works.)

(Just for reference, there are examples in:

 * https://github.com/serilog/serilog-sinks-splunk/
 * https://github.com/serilog/serilog-sinks-seq/

both showing how the packaging/build works in `dotnet` for similar sinks.)

I am happy to help out with this, I have hesitated so far because I don't have much knowledge of the plumbing of ES, nor an ES instance to test with, so I'm probably not the best one to do it....

Here's a plan that might work:

 1. As in the other repos, `dev` can become the "v2/.NET Core" branch here
 2. I'll merge this PR manually to my local `dev`, to take advantage of @alexvaluyskiy's work so far, and then update what I can to RC2 conventions
 3. I'll push the result, even a partially broken one, to `dev`
 4. We can hack at it/PR into `dev` until everything's fixed (an opportunity for anyone else blocked at this for RC2 to get involved)

Workable? Is there a better plan? :-)

gdoron:
@nblumhardt I can probably  donate an elastic search instance for this cause 😉.
I can start a dedicated instance on azure for that.
Though, AFAIK there's no difference between a local instance and a remote one, you can simply test with an elastic search instance on your working machine.

mivano:
I m all in favour of adding a .NET Core version, but I want to avoid dropping support for the normal framework. There are still improvements coming in, if they only come to the .NET Core package then we miss out. At my work for example we are not ready yet to go to .NET Core. So if it is possible to pack both .NET Core and other versions into the same nuget package, then certainly yes. Otherwise I believe we need to offer a second package.

gdoron:
@mivano Yes, we do need to realize that though .NET Core is the hottest thing around and new kid in the block, its market share is not even 0.1% (a made up number I can't prove :laughing: ) of the industry.

nblumhardt:
@mivano https://github.com/serilog/serilog/blob/dev/src/Serilog/project.json#L15 - yep! :-) ... multi-targeting is one of the big-ticket features of the new tooling.

Before I plunge on though, is there really support for this? I can get the basics merged, but it's going to require some tweaking/testing/proving before it'll be usable, and I'm not able to do that right now. If no one is keen I'd rather save the effort...

Is someone with ES skills, and familiarity with the way the sink works, able to take point on getting the end result across the line?

Cheers!

Mpdreamz:
So this PR was for RC1 of dotnet core using the DNX toolchain. 

Thats the version with a go live license and the one that the published version of `Elasticsearch.Net` on nuget.org supports. 

[We have a PR for RC2 support](https://github.com/elastic/elasticsearch-net/pull/2111) but I was not planning to release that as part of our `2.x` releases, only the the `5.x` pre-releases.

@nblumhardt are you betting completely on RC2 even with the state of the `dotnet` tooling in such a flux? I'm personally very hesitant to rely on the `dotnet` tooling at all in my builds. 

There is no `dnvm` equivalent so there's no way to be sure everyone is using the same `dotnet` tooling and thus have repeatable builds. I opened a ticket [for this in january](https://github.com/dotnet/cli/issues/1022) but not much has happened there since and is still currently on the backlog.

I feel bad enough my `2.x` releases are reliant on the `dnvm` `dnx` tooling but atleast those have a go live license and [atleast we can check and download the expected versions automatically in our build](https://github.com/elastic/elasticsearch-net/blob/master/build/scripts/Paths.fsx#L261-L264). 

Even so we are moving back to a dual build for the full framework (and PCL's) through msbuild (paket is a huge huge help for multitargeting platforms already) and only build and test the dotnet core build using the `dnx` / `dotnet` tooling. 

If the version manager story is not available by the time we are ready to release our `5.x` version, i'd personally rather rollback and release our `5.x` only with `RC1` support while the dust settles and we get the next go live release (`RC2-final` or `Preview 1`) for coreclr. 

That said i'm personally not ready to support this sink for RC2 while everything (project system, tooling) is in such a flux.

nblumhardt:
@Mpdreamz yes, the tooling situation is a sub-optimal state of affairs, I agree entirely. I have a bit of _pre-release package fatigue_ at the moment too, I hear where you're coming from :-)

I've had a few people reach out on various channels about getting Elasticsearch supported on RC2, but like you say, there's a bit of effort involved in chasing the moving target and unclear adoption at this point. Looking at the [download counts for the Splunk sink](https://www.nuget.org/packages/serilog.sinks.splunk) it seems like about 20% since the 18th of May have been the RC2 version, so there's some interest out there, but I'm not sure what this means in reality.

It sounds like the discussion is moot until the Elasticsearch.NET PR lands; once it does, maybe we widen the net? So long as the unstable work is going on in `dev`, with `master` clean and tidy, perhaps we can call it "best effort"? I'm happy to do the initial merge and try to get things building; over time with some interest, perhaps the wider community can iron any wrinkles out.

Not sure if it addresses the lack of `dnvm`, but the `dotnet-install.ps1` script called from `appveyor.yml` allows us to specify a concrete `dotnet` version for the builds of the other Serilog sinks. May be a partial solution, anyway?

Cheers!

Mpdreamz:
> . I have a bit of pre-release package fatigue at the moment too.

I'm stealing this phrase from you! :smile: I didn't mean to stop this entirely in the tracks just cause I'm suffering from the same fatigue. 

Plan of attack proposal:

We're going to pull in rc2 support for the es client next week but it will only make its way into our 5.x prereleases for now. Hinging on that is some build infrastructure refactoring where in we do a dual build using both `msbuild` and `dotnet`. We do this simply because we might build using `dotnet 1.0.0-preview2-002823` today but who knows where that download will be in 1-2 years from now. At least this way repeatable builds for the full .net frameworks are not in danger.

The 5.x version of the low level client this sink uses is only tested against ES 5.x, I'm going to create a test suite specifically for the API's used in this sink to make sure they are continuously tested against a 2.x elasticsearch instance too. The ES client had quite a few breaking changes going from `1.x` to `2.x`, we plan to keep that at a minimum for 2.x to 5.x upgrade. Ideally the next update of the sink will work against both Elasticsearch 2.x and 5.x.

If you are OK prepping this sink over to rc2 hopefully by the end of next week we can both do another prerelease to either add to or alleviate our prerelease fatigue :smile:

nblumhardt:
:smile: sounds like a plan :+1: ... if you can ping this thread when the 5.x package is available, I'll merge/set up the build then.

PR #22, opened 2016-02-03
fix issue with ElasticSearch not supporting '.' in fieldnames

leachdaniel:
Dots are not allowed in Field Names, replaces '.' with '/' https://github.com/elastic/elasticsearch/issues/14594

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
@@ -49,9 +49,11 @@
     <Reference Include="Elasticsearch.Net">
       <HintPath>..\..\packages\Elasticsearch.Net.1.7.1\lib\net45\Elasticsearch.Net.dll</HintPath>
     </Reference>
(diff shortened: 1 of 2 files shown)


mivano:
Thanks Daniel for the PR. I was not aware of this, but looking at the link you provided it looks like a good solution. @nblumhardt or @Mpdreamz you see any drawbacks of this? Thinking if this breaks anything...

nblumhardt:
Looks okay to me, FWIW (but not familiar with the issue).

Mpdreamz:
This looks good to me :+1: field names no longer can contain dots because it causes a lot of ambiguity when resolving `deeply.nested.fields.in.json.graphs` 

This might break some folks kibana dashboard but fixing this at an elasticsearch level for older indices is also mandatory when upgrading to Elasticsearch 2.0

https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_mapping_changes.html#_field_names_may_not_contain_dots

this plugin https://github.com/elastic/elasticsearch-migration is a godsend if you are planning to upgrade from Elasticsearch 1.x to 2.x. 

@mivano pulling this and #23 warrants a major version imo.

mivano:
Agreed, lets pull this into the 2.0 NEST branch and bump the major. I will try to have a look at it soon.

mivano:
Merged in the new v3 version. Thanks for the PR!

PR #23, opened 2016-02-15
Updated ES sink for compatibility with Elasticsearch.Net and NEST 2.0…

konste:
…2. Elasticsearch got many breaking changes, so update of the sink was not strightforward. My biggest area of concern is ElasticsearchSinkTestBase.cs - I think I shortcut some tests, because underlying interfaces got cut dramatically.

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch-net40.csproj
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch-net40.csproj
@@ -46,8 +46,9 @@
     <AssemblyOriginatorKeyFile>..\..\assets\Serilog.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Elasticsearch.Net">
-      <HintPath>..\..\packages\Elasticsearch.Net.1.7.1\lib\net40\Elasticsearch.Net.dll</HintPath>
+    <Reference Include="Elasticsearch.Net, Version=2.0.0.0, Culture=neutral, PublicKeyToken=96c599bbe3e70f5d, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Elasticsearch.Net.2.0.2\lib\net45\Elasticsearch.Net.dll</HintPath>
+      <Private>True</Private>
     </Reference>
     <Reference Include="Serilog">
       <HintPath>..\..\packages\Serilog.1.5.14\lib\net40\Serilog.dll</HintPath>
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
@@ -46,8 +46,9 @@
     <AssemblyOriginatorKeyFile>..\..\assets\Serilog.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Elasticsearch.Net">
-      <HintPath>..\..\packages\Elasticsearch.Net.1.7.1\lib\net45\Elasticsearch.Net.dll</HintPath>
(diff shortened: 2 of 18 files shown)


Mpdreamz:
Quickly skimming the PR the changes look good to me :+1: . I'll be out for the next 9 days so my chances of reviewing this by pulling down and testing then are slim.

konste:
I will certainly work on ironing out automated checks. Would anybody be able to review and pull it in sooner?

konste:
Code seems to be fine now, but tests are failing, because I don't know how to implement ElasticsearchSinkTestsBase constructor properly. That FakeItEasy thing seriously f...ake my brain and I does not seem to be able to fix it in 3 hours I have allotted to it.

tabro:
@konste Still looking into this one? I am having issues with not being able to log on my ES 2 cluster, so I am eagerly awaiting this PR :) Let me know if I can help out with something.

konste:
@tabro as far as I understand the version of the sink in my branch compiles and works. It fails 6 tests though because I believe actual behavior of Elasticsearch.Net component changed and those tests need to be corrected. It would help if you try to iron out those tests, so @Mpdreamz can integrate it when he is back.

tabro:
@konste I have a commit that fixes the unittests - but I cant commit to your fork repo - can you add me as contributer?

tabro:
@konste Nevermind that - I sent you a Pull Request on your fork.

konste:
Hurray! @Mpdreamz, waiting on you now.

mivano:
Nice job @konste and @tabro. It all looks fine to me so I m more than happy to proceed. Do we want to bump this to a new major version? So how breaking is this? Can we still use this against 1.x versions of ES? Any recommendations?

We also need to make sure we update the readme and changelog for this.

konste:
We CANNOT use it with ES 1.x, so it is very much breaking. We just follow the pattern of Elasticsearch.NET itself. So it definitely makes sense to make it new major version.

mivano:
Cool, so we need to have two versions out in the wild then. @nblumhardt is this possible with appveyor? Have similar things in other sinks?

As it is breaking already, would it be sense to include PR #22 into this too?

megakid:
Any news on this? We are running elastic 2.0 and really want to use this sink!

mivano:
I really want to bring this forward, but we need to support both a 1.x version and a 2.x version. And I have no clue how we can configure this in the build system. @nblumhardt any help/pointers?

nblumhardt:
It doesn't seem so much a build issue but of package naming/versioning. I'd suggest bumping the major version to 3.0 and moving on: just like with other platform shifts, users of the (sink version) 2.x can always depend on the older package. We could branch off `maintenance-v2` in this repo to support patches on the older version if/when necessary.

If that isn't viable here, then creating a new package with a new name for ES 2.0 is the only option I can think of that's NuGet compatible. Doing this incurs some support/documentation debt that will be around for a long time, however.

Not sure of the background here so I may of course be missing the point :-)

mivano:
Agreed, lets bump it to v3, document the change, keep an old version branch and if needed create a different ES v1 package for it. I m on it.

konste:
Thank you for doing that! I need ES 2.0 compatible client too, but I don't feel comfortable to deal with all that build infrastructure.

nblumhardt:
@mivano here's an _appveyor.yml_ file that can be included in the root of the repo to control CI. This is (roughly) what the current build uses - including this and modifying it will modify CI.

```yml
version: '{build}'
build_script:
- ps: ./Build.ps1 -majorMinor "2.0" -patch "$env:APPVEYOR_BUILD_VERSION" -customLogger "C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
artifacts:
- path: Serilog.Sinks.Elasticsearch*.nupkg
deploy:
- provider: NuGet
  api_key:
    secure: nvZ/z+pMS91b3kG4DgfES5AcmwwGoBYQxr9kp4XiJHj25SAlgdIxFx++1N0lFH2x
  on:
    branch: master
```

mivano:
Thanks, will include it. Updating readme and changelog now.

mivano:
Done, package created. However, it continued counting from the last 2.0 version, so the package version is now 3.0.96 instead. Resetting it to one has no use unless we can remove the existing builds.

PR #31, opened 2016-03-02
Added back off capability to the durable shipper

mivano:
As described in this issue https://github.com/serilog/serilog-sinks-seq/pull/13
This PR adds a back off mechanism to the durable log shipper.


nblumhardt:
LGTM, will let you know if any issues show up via the Seq impl.

PR #32, opened 2016-03-02
Removed .NET 4 support as Elasticsearch.NET is not supporting it anym…

mivano:
At the moment we do have support for .net 4 besides the 4.5 assemblies we create. This means we need to maintain an extra project file, solution file and more complex code. As the Elasticsearch.NET 2 assembly no longer has .NET 4 support, it is no longer needed to support it in the v3 version of this sink. 

This PR removes the dependencies for .NET 4.


Mpdreamz:
Lgtm :)

konste:
LGTM. The only big thing I really miss in this sink is durable implementation derived from non-durable, rather than being standalone.

mivano:
Cool, thanks. Updated the documentation and merged.

mivano:
The durable section is borrowed from the seq client sink and I get your point. This sink has a lot of good features to get pumping data into ES quickly. But if real durability is needed, then a queue or file based system in combination with Logstash might be a better solution. More work to set up maybe, so that is why this sink is ideal :-)

PR #34, opened 2016-03-07
Replaced DynamicResponse with BulkResponse

konste:
This is the change which motivated me to update ES sink for NEST 2.0 compatibility. EmitBatchChecked return type is not very generic and hard to handle DynamicResponse, but rather very specific and much less error prone to handle BulkResponse. Result of ES bulk indexing request can be checked for success (and errors may be reported otherwise) much faster and easier than before.


konste:
I'm wrestling with Git, and so far it wins :-(

PR #35, opened 2016-03-07
Replaced DynamicResponse with BulkResponse

konste:
This is the change which motivated me to update ES sink for NEST 2.0 compatibility. EmitBatchChecked return type is not very generic and hard to handle DynamicResponse, but rather very specific and much less error prone to handle BulkResponse. Result of ES bulk indexing request can be checked for success (and errors may be reported otherwise) much faster and easier than before.

This time net40 is not forgotten and all tests are passing!

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch-net40.csproj
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch-net40.csproj
@@ -50,6 +50,14 @@
       <HintPath>..\..\packages\Elasticsearch.Net.2.0.2\lib\net45\Elasticsearch.Net.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="Nest, Version=2.0.0.0, Culture=neutral, PublicKeyToken=96c599bbe3e70f5d, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\NEST.2.0.4\lib\net45\Nest.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Newtonsoft.Json.8.0.2\lib\net40\Newtonsoft.Json.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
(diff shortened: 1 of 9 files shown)


Mpdreamz:
What I personally don't like is that this introduces a Json.NET transient dependency by including NEST. 

We could include a dedicated `BulkResponse` class instead of relying on `DynamicResponse`. The low level client also supports deserialization to types out of the box albeit using SimpleJSON (internalized dep of ES.NET).

konste:
How about that:
``

       /// <summary>
        /// Emit a batch of log events, running to completion synchronously.
        /// </summary>
        /// <param name="events">The events to emit.</param>
        /// <returns>Response from Elasticsearch</returns>
        protected virtual ElasticsearchResponse<T> EmitBatchChecked<T>(IEnumerable<LogEvent> events) where T : class
        {
            // ReSharper disable PossibleMultipleEnumeration
            if (events == null || !events.Any())
                return null;

            var payload = new List<string>();
            foreach (var e in events)
            {
                var indexName = _state.GetIndexForEvent(e, e.Timestamp.ToUniversalTime());
                var action = new { index = new { _index = indexName, _type = _state.Options.TypeName } };
                var actionJson = _state.Serialize(action);
                payload.Add(actionJson);
                var sw = new StringWriter();
                _state.Formatter.Format(e, sw);
                payload.Add(sw.ToString());
            }
            return _state.Client.Bulk<T>(payload);
        }

konste:
Those who already have Jason.NET in their projects can use BulkResponse. Those who don't my continue to use DynamicResponse.

mivano:
I liked the fact that we did use the low level elasticsearch.net library only. Introducing NEST and JSON.NET for this add extra weight. So I rather prefer your other solution then.

Furthermore; I really think we should drop .NET 4 support since Elasticsearch.net even does not support it any more. See the PR #32 for this. 

I also think we can use dynamics for it and provide something like this: https://github.com/serilog/serilog-sinks-elasticsearch/blob/handlefailures/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs#L60

Would that not suffice too?

konste:
I agree with dropping .NET 4 any moment - please do!

Regarding our original topic - people may have very different strategy for handling ES request failures. For my purposes using standard test stream based SelfLog is not sufficient. I use more complex logger and more specialized strategy to handle ES failures.

Based on that I'm going to close this PR - also don't want to add unnecessary weight. 
Instead I will open another PR with generics based solution shortly.

PR #39, opened 2016-03-09
Improved handle failures

megakid:
Small improvements to the handle failure cases.  Untested, I will take a closer look soon.


mivano:
See #127.

PR #41, opened 2016-03-26
Add KibanaFriendlyJsonFormatter

### LINKED ISSUE #37: Exceptions as an array breaks Kibana usage
A recent commit https://github.com/serilog/serilog-sinks-elasticsearch/commit/010af25d1641b009afbe47b15bc46fc7b4a070f9 changes the behaviour from writing exceptions to ES from writing an object in the "exception" field into writing an array of exceptions in "exceptions".

This breaks Kibana dashboards based on the ES logging, since Kibana doesn't play well with arrays. For instance we can't put exception.message on a dashboard.

While I understand the reason for making the change is probably to avoid deep nesting in the case of innerExceptions, I think it'd be beneficial to keep a single exception object in the "exception" field.

I'd love to provide a pull request for this, if we can agree on what behaviour it should have ? Perhaps make it configurable so users can configure the sink to use the old or new behaviour ?

driis:
This PR should fix #37 by providing users an alternative formatter to pass in as the custom formatter. This will restore pre-2.0 exception formatting; thus avoiding putting exceptions in an array which creates problems when using the sink with Kibana.

To use it, users should pass in the new KibanaFriendlyJsonFormatter in options:

```
var sink = new ElasticsearchSink(
  new ElasticsearchSinkOptions(uri) 
  { 
    CustomFormatter = new KibanaFriendlyJsonFormatter(renderMessage:true)
  });
```

Default behaviour has not been changed.

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
@@ -50,6 +50,10 @@
       <HintPath>..\..\packages\Elasticsearch.Net.2.0.4\lib\net45\Elasticsearch.Net.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="Newtonsoft.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\Newtonsoft.Json.8.0.2\lib\net45\Newtonsoft.Json.dll</HintPath>
+    </Reference>
     <Reference Include="Serilog">
       <HintPath>..\..\packages\Serilog.1.5.14\lib\net45\Serilog.dll</HintPath>
       <SpecificVersion>False</SpecificVersion>
@@ -76,6 +80,7 @@
     <Compile Include="Sinks\Elasticsearch\ElasticsearchSinkOptions.cs" />
     <Compile Include="Sinks\Elasticsearch\ElasticsearchSinkState.cs" />
     <Compile Include="Sinks\Elasticsearch\ExponentialBackoffConnectionSchedule.cs" />
+    <Compile Include="Sinks\Elasticsearch\KibanaFriendlyJsonFormatter.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="..\..\assets\Serilog.snk">
--- a/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
+++ b/src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
@@ -50,10 +50,6 @@
       <HintPath>..\..\packages\Elasticsearch.Net.2.0.4\lib\net45\Elasticsearch.Net.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="Newtonsoft.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\..\packages\Newtonsoft.Json.8.0.2\lib\net45\Newtonsoft.Json.dll</HintPath>
-    </Reference>
     <Reference Include="Serilog">
       <HintPath>..\..\packages\Serilog.1.5.14\lib\net45\Serilog.dll</HintPath>
       <SpecificVersion>False</SpecificVersion>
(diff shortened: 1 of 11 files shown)


src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
```diff
@@ -50,6 +50,10 @@
       <HintPath>..\..\packages\Elasticsearch.Net.2.0.4\lib\net45\Elasticsearch.Net.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="Newtonsoft.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
```
mivano: This adds an additional dependency on newtonsoft.json (and jt looks like it is missing in the nuspec file). I m not a fan of adding these core frameworks as a reference here. The cleaner, the better. Is this really needed?

mivano:
Thanks for the PR. I see the purpose, but the name is a bit misleading. The current serializer is also fine for Kibana so another name might be more clear.

src/Serilog.Sinks.Elasticsearch/Serilog.Sinks.Elasticsearch.csproj
driis: It's an error that this was included. Fixed in 87ba8e4d3dc10c0e5c6b6ac0bc43d81c5c2cda63

driis:
@mivano thanks, I can agree to the misleading name. Naming is hard.

Do you have any naming preference ? Off the bat I'd suggest a rename of the new formatter to `ExceptionAsObjectJsonFormatter`, perhaps ?

mivano:
By lack of any better idea I think it is the best suited name. Can you also include some description in the readme.md and changelog.md files?

driis:
Rename done, added a note to changelog and README.md.

Mpdreamz:
The initial idea behind #37 flattening was to avoid creating too many Elasticsearch fields e.g exception.innerException.innerException.......message

And to include the top level exception under `exception` without innerException

We opted out of the latter to not introduce duplication

I'd rather we (potentially configurably) reintroduce this on the default formatter if it simplifies some kibana widgets rather then adding another formatter. 

Sorry for the late reply!

driis:
@Mpdreamz I understand that reasoning. The reason I thought this design was a good idea, is that it's a clean way to give users the option of getting the old behaviour back, without polluting the current formatter.

I could do the first exception in the `exception` field as a new pull request, if you'd like. In that case,

* Should it be on by default, and how should users opt-in ? I'm still thinking a new formatter, because it's clean given the current design.
* We could avoid duplication by renaming `exceptions` to `innerExceptions` and _not_ include the top level exception there. That would be a breaking change, however. Again, if we do it as opt-in, that should not be a problem.

Mpdreamz:
I don't see adding this as an option to the sink as polluting since this scratches as itch everyone might have :smile:

I almost rather inverse this PR, log `exception` by default and ship a formatter/option that does not.

In both cases we also need to update the template so that `exception` is explicitly mapped.

* Opt in / Opt out, this is tough because for someone we'll always choose the wrong default. I'm inclined to do opt out, I rather have the information there easily accessible in all corners of KIbana for newcomers

* That would make querying much harder though, i'm fine with the duplication. `_source` is compressed and the increase in index size will be manageable. The main "pain" in duplication comes from introducing more fields not duplicating the same data across multiple fields in this case.

Sorry if I derail the PR too much, @mivano obviously has final say on the direction here :+1:

mivano:
Good arguments, thanks. For me the missing top exception and having an array of exceptions is not an issue. Normally some part of the error is inside the message already and we use Kibana and expand the details anyway to see the full list of exceptions and details for each of them. So an opt in would be okay for me and just switching in another formatter makes this easy.

Benefit is also that this is not again a change like we had from version 1 to 2, does not introduce all the fields and template change and will still allow people to switch to other kinds of behaviour. Drawback can indeed be the lack of direct visibility in Kibana (but have not seen a lot of requests for this) and the fact that it is less obvious to configure the other behaviour when using for example the appSettings configuration.

I would still go for the suggested PR; a custom formatter, documentation and opt in. If more requests come in to make this the default, we can swap out the formatters or make it configurable in another way.
I would want to give this sink some stability now as we already bumped and introduced some changes recently. 

But lets leave this open for some days so others can share their opinions.

driis:
I agree that another breaking change on the sink right now is probably not a good idea. With adequate documentation the new formatter should be discoverable to those who need it. 

Can we reach a consensus here ? If we go with the new formatter, is there anything you need me to add to the PR ?

Mpdreamz:
:+1: sorry for derailing :)

mivano:
Sorry @driis, I did not had a lot of time yet. The PR looks okay to me, so I will try to pull it in soon, most likely tomorrow.

mivano:
Merged, packages should be out soon. Thanks for the PR!

guillaume86:
@mivano, regarding your comment;

> Thanks for the PR. I see the purpose, but the name is a bit misleading. The current serializer is also fine for Kibana so another name might be more clear.

Could you give some details on how you use kibana with exceptions inside an object array? 
All I managed to do is viewing the `exceptions` field as a big json blob in the discover view, but that's pretty much it, I can't search/filter them, can't build graphs, etc

PR #46, opened 2016-06-13
Added possibility to override level names

skarum:
Override log level names when serializing to match naming of other frameworks.

E.g. to match Bunyon (https://github.com/trentm/node-bunyan) Information should be logged as info.

This PR makes it possible to specify a text value for each log level.


mivano:
Thanks for the PR. It looks to me like a very specific use case. I see you make it all optional so it should not hurt, but the majority of the change is in the Json formatter. Would a custom formatter that you can insert not cover your scenario already?

skarum:
Ohh! You are totally right, don't know why I didn't think about some like that. 

It's a quite specific need and overriding the ````WriteLevel(LogEventLevel level, ref string delim, TextWriter output)```` method is much cleaner than injecting the Dictionary into the formatter.

Thank you for the response.

mivano:
Cool, if that solves your need, then I rather do not add this specific case into the sink directly and I will close this PR. Thanks!

PR #51, opened 2016-06-26
Add options to provide a custom template and overwrite an existing

blachniet:
These changes allow the user to supply a custom template and overwrite the template if it already exists in Elasticsearch.

`GetTemplateContent` can be set to return the body of the PUT template HTTP request. This allows users to further customize the Elasticsearch template, instead of using the standard Serilog template.

If `OverwriteTemplate` is set to true, the state class will not check to see if the template already exists before PUT-ing the new one, which will overwrite the template if it already exists.


blachniet:
Anything I can do to help get these changes, or something like them, integrated in? I've hacked around with my project to get this sort of functionality in, but I think it would make sense to have it in the sink. Are there oppositions to the changes I made, or something that I just screwed up on? Thanks!

nblumhardt:
Hi @blachniet,

I'm not an active maintainer of this sink but I notice your initial PR commit was non-mergeable and so CI failed. Merging upstream/dev back into your PR branch should kick off a new build.

Hope this helps,
Nick

mivano:
Cool! Can you also add some details to the readme.md file?

src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkOptions.cs
```diff
@@ -41,6 +41,18 @@ public class ElasticsearchSinkOptions
         /// </summary>
         public string TemplateName { get; set; }
 
+        /// <summary>
+        /// When using the <see cref="AutoRegisterTemplate"/> feature, this allows you to override the default template content.
+        /// If not provided, a default template that is optimized to deal with Serilog events is used.
+        /// </summary>
+        public Func<string> GetTemplateContent { get; set; }
```
Mpdreamz: It might be more useful to return `object` here, the implementer could return a string or anonymous object in that case.

blachniet:
I made some of the suggested changes:
* Merged in latest dev branch
* Changed `GetTemplateContent` to return an `object` instead of `string`
* Mentioned ability to define a custom template mapping in the README

I noticed that most of the code examples and feature details are described in wiki pages as opposed to the README. Should I add a code example there, with a link to it from the README, after it's merged in? Or should I add all that to the README?

Thanks!

mivano:
Thanks @blachniet for the PR! I merged it now to dev and will create a merge to master soon to create new packages. 

Regarding the wiki; I notice that I rather maintain the readme file then the wiki but they need to be in sync somehow. I will update it when it is merged to master.

PR #58, opened 2016-08-04
Break Serilog.Formatting.Json.JsonFormatter dependency (WIP)

nblumhardt:
Relates to serilog/serilog#819.

I thought I'd put this up in it's current state for discussion - this PR just moves the base JSON formatter type into the sink assembly as `DefaultJsonFormatter`.

There are a couple of options for how this might be cleaned up - a zero-breaking-change option, if that's desired, is to merge as-is.

Another alternative is to get rid of some of the unnecessary cruft, e.g. `omitEnclosingObject` and so-on, and perhaps collapse `DefaultJsonFormatter` down into `ElasticsearchJsonFormatter`, simplifying and cleaning things up in the process.

Happy to try either approach, keen to gather some feedback @mivano @Mpdreamz @merbla.


Mpdreamz:
LGTM :+1: its nice to add the non breaking variant in 4. 

When we accumulate enough breaking changes moving over to `one true new way` should be a blocker in 5.

nblumhardt:
Thanks @Mpdreamz.

:shipit: ?

nblumhardt:
Guys, closing this to re-target to `dev`. Cheers!

PR #59, opened 2016-08-04
.NET core support!

Mpdreamz:
Lotsa changes, #21  was still super useful as a reference here (thank you @alexvaluyskiy!)

Also moved the test project over to dotnetcore and xunit, nunit has a dotnet runner but it was throwing hard exceptions and only testing one framework. The xunit dotnet runner tests both full desktop clr as well as netstandard1.3 (core clr). 

Fixed tests due to 2.0 changes, the first emit() now always sends directly irregardless of `BatchPostingLimit` or period. The different timer implementations also do not guarantee that this first emit() will only send one document over. 

Super helpful to have @nblumhardt and @merbla in person at NDC to help out :D cheers guys!

@mivano I ported over the powershell build script as far as I could as I'm not sure how you do releasing and appveyor so this needs another pass over to suit your deployment model. 

It builds and tests.  It also packs nuget packages in the same location but please double check the dlls have the correct version applied, the nuget metadata is valid. same nuget id with casing, are the dlls signed as we expect them too?


src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/CrossPlatform/PortableTimer.cs
```diff
@@ -0,0 +1,113 @@
...
+// limitations under the License.
+
+#if NO_TIMER
+
+using Serilog.Debugging;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Serilog.Sinks.Elasticsearch.Sinks.ElasticSearch.CrossPlatform
```
nblumhardt: Namespace is a bit crazy here, should be `Serilog.Sinks.Elasticsearch.CrossPlatform`.

src/Serilog.Sinks.Elasticsearch/project.json
```diff
@@ -0,0 +1,61 @@
+{
+    "version": "4.1.0",
```
nblumhardt: For the new build setup to work this needs to be `4.1.0-*` - the scripts will set the suffix correctly to nothing (the `-` disappears) or `-branch-nnnnnn` depending on whether or not `master` is being built.

nblumhardt:
LGTM! :+1: Added some very minor notes.

Not related directly to the PR, but what do you think about making `dev` the default branch in this repository? We've switched from `master` to `dev` in most/all of the other repos so that new work is always visible (makes liveness obvious) and so that new PRs will target it by default. Just a thought.

mivano:
Thanks guys! Looking good. I picked the items from @nblumhardt but the feature branch building does not work correct with the 4.1.0-* trick. Getting this now:

4.1.0-feature/do-00152' is not a valid version string.

It looks like it is the same approach as used with Serilog itself, so maybe @nblumhardt has a tip here? We need to extend this to handle feature branches named feature/?

nblumhardt:
@mivano ah, I see - you have a branch name with a slash in it... I think it'd be easy enough to replace any instances of `/` in the branch name with `-` and get the script working, we've only not accounted for that case because it hasn't been an issue in the other repositories (yet) :-)

mivano:
I 'fixed' the versioning by using gitversion. That gives better versions per branch.

mivano:
@nblumhardt I agree that switching to dev as default branch makes sense. I will make the switch.

Mpdreamz:
+1 on gitversion solves two problems in one go :), pitty it needs to change the working directory but I guess failures here leaving you in a diff older as the one we started should be super rare.

mivano:
I m a bit new to dotnetcore so it can most likely be implemented in a better way, but this made sure you could run it both locally and on the build server.

mivano:
Took some time, but merged to master and it has packages now: https://www.nuget.org/packages/Serilog.Sinks.Elasticsearch

Thanks for the all the great work!

Mpdreamz:
Awesome @mivano thanks for taking this PR to 100% :tada: :balloon: :fireworks: :cake: :ship: :it:

PR #61, opened 2016-08-08
Pull in Serilog's default JSON formatter

nblumhardt:
...to break dependency on Serilog.Formatting.Json.JsonFormatter virtual extension points (serilog/serilog#819)

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/DefaultJsonFormatter.cs
+++ b/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/DefaultJsonFormatter.cs
@@ -1,0 +1,396 @@
+﻿// Copyright 2013-2016 Serilog Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


nblumhardt:
```
The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: Unhandled Exception: GitTools.WarningException: It looks like the branch being examined is a detached Head pointing to commit 'acafa4f'. Without a proper branch name GitVersion cannot determine the build version.
```

It looks like some more work may be required for GitVersion to work for PRs, or perhaps it could just default to some fixed version when in detached head state? CC @mivano

nblumhardt:
Any guidance on this one folks? Never having used GitVersion I'm not sure where to start; would using the versioning scheme from the other Serilog repos help here? (I can PR if so).

mivano:
Hmm that is a nice one. We are using gitversion with TFS builds and never run into this. For PRs (feature branches) it creates nice x.y.z-branchname0001 versions. Which are nuget compatible. Are you aware of an option in Appveyor to tell it to get the branch by its name? Otherwise we need to make it optional for PRs and just use a random version number. But that defeats the nice gitversion way...

mivano:
I checked in to dev an updated build.ps1 file that ignores gitversion when on a head less branch. Can you try and see if this fixes the issue?

nblumhardt:
@mivano :white_check_mark: ready to go!

mivano:
Great, I merged it back to dev.

nblumhardt:
Thanks :+1:

PR #64, opened 2016-08-24
Backport exponential backoff into v2

amar-tcpl:
Unfortunately we're stuck on ES 1.5 so have to stick with the 2.x clients, however with our production workloads the exponential backoff functionality is very important for when we need to perform cluster maintenance. I've backported the code from 4.x and everything works great.

I've created a temporary package in nuget for anyone who wants this feature today:
https://www.nuget.org/packages/Serilog.Sinks.Elasticsearch.Backoff/2.2.0

Thanks,


mivano:
Thanks! I merged it in the maintenance branch. Be aware that there is no build/nuget package from that branch. So thanks for offering your package from Nuget.

PR #77, opened 2016-12-13
Update README.md

mattlaver:
Update documentation to use .SetBasePath when configuring appsettings.json to avoid System.IO.FileNotFoundException.


mivano:
Thanks!

PR #82, opened 2017-01-07
More options in config file.

StevenVandenBroeck:


mivano:
Thanks for the PR. Makes sense to support more parameters from the config. However, by adding them in the extension as optional parameters with default values, do you not introduce a breaking change? Would adding a new overloaded method be a better option?

StevenVandenBroeck:
I've used the same defaults as in ElasticsearchSinkOptions, but I can change it to an overload.

StevenVandenBroeck:
I can't make an overload because of the optional parameters (ambiguous error).

mivano:
Ok, then we use the optional parameters. Thanks for the PR, I will merge it to dev.

Issue #87, opened 2017-01-25
specify Authorization header in configuration


AmirSasson:
i am using the Serilog.Configuration to configure my serilog, and i am writing my logs to Elastic search.
my ElasticSearch instance requires  Bearer Token in the Authorization Header.
at the moment there is no way specifying the Authorization header in the web/app.config file.
only pragmatically by specifying : ModifyConnectionSettings.

mivano:
Correct, there is no overload that you can use for this as it requires some pretty specific code to support this authentication. Do you have an solution in mind?

AmirSasson:
maybe implement something that will support
`<add key="serilog:write-to:Elasticsearch.connectionGlobalHeaders" value="Authorization=SOMETOKEN&SOME-OTHER-HEADER=SOME-OTHER-VALUE" />`

deserialize the the **value** into key value pair and add global headers this way:
 ModifyConnectionSettings = (c) => c.GlobalHeaders(new NameValueCollection { { "Authorization", config.ESLogsAuthToken } })

does this make sense?

Issue #75, opened 2016-11-15
Add support of basic authentication


agamanyuk:

mivano:
Would be a good addition. You can now use the Connection property to have control over the way the connection is set up and use that to pass in your credentials.

AmirSasson:
you can achieve this by doing:
``` 
            ElasticsearchSinkOptions options = new ElasticsearchSinkOptions(new Uri(config.ElasticUri))
            {
                IndexFormat = $"SOME-FORMAT-{0:yyyy.MM}",
                AutoRegisterTemplate = true,
                ModifyConnectionSettings = (c) => c.GlobalHeaders(new NameValueCollection { { "Authorization", "Basic TOKEN-TOKEN-TOKEN-TOKEN-" } })
            };


            loggerConfig.WriteTo.Elasticsearch(options);
```

megakid:
Or if you want default windows credentials like we use:

    public class UseDefaultCredentialsHttpConnection : HttpConnection
    {
        protected override void SetBasicAuthenticationIfNeeded(HttpWebRequest request, RequestData requestData)
        {
            // Do nothing
        }

        protected override HttpWebRequest CreateWebRequest(RequestData requestData)
        {
            var request = base.CreateWebRequest(requestData);

            // Enable kerberos auths 
            request.PreAuthenticate = true;
            request.UseDefaultCredentials = true;

            return request;
        }
    }

    var elasticOptions = new ElasticsearchSinkOptions(_uri)
    {
        Connection = new UseDefaultCredentialsHttpConnection(),
        ModifyConnectionSettings = c => c.EnableHttpCompression(), // Not needed for Auth but shows what else can be done
    };

PR #90, opened 2017-01-26
#87 Add Support for specifying Http Headers in config

AmirSasson:
#87 developers can specify http headers in the config file like:
<add key="serilog:write-to:Elasticsearch.connectionGlobalHeaders" value="Authorization=Bearer SOME-TOKEN;OtherHeader=OTHER-HEADER-VALUE" />
for elastic servers that require Authorization header.
the parameter is not mandatory.

this solves also issue #75

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ This example shows the options that are currently available when using the appSe
     <add key="serilog:write-to:Elasticsearch.minimumLogEventLevel" value="Warning"/>
     <add key="serilog:write-to:Elasticsearch.bufferBaseFilename" value="C:\Temp\SerilogElasticBuffer"/>
     <add key="serilog:write-to:Elasticsearch.bufferFileSizeLimitBytes" value="5242880"/>
-    <add key="serilog:write-to:Elasticsearch.bufferLogShippingInterval" value="5000"/>
+    <add key="serilog:write-to:Elasticsearch.bufferLogShippingInterval" value="5000"/>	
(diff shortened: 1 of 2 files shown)


AmirSasson:
@mivano, are you going to pull this request?
this is a blocker for us. we write to elastic search only on production which means we must use Serilog app settings/configuration. and our elasticSearch server requires some authorization (like most elasticSearch  out there..).
since there is no way to specify authorization in config -> this is a blocker. and this pull request overcomes the problem.
it support all kind of authorizations: Bearer, basic,... whatever..
is there a chance you can respond?
tx

mivano:
Sorry for the slow response. I hope I can dive into this soon. Do consider that you can add your own connection builder and read from the appSettings the right config values.

mivano:
The 4.2 package is available. thanks for the PR!

CumpsD:
https://github.com/serilog/serilog-sinks-elasticsearch/blob/dev/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs#L145

Currently it splits on `=`, it is possibly to have a value with an equal sign in it (base 64 encoded values), shouldn't we do:

```
var headerValue = headerValueStr.Split(new[] { '=' }, 2, StringSplitOptions.RemoveEmptyEntries);
```

AmirSasson:
@CumpsD, sure, we can change this to limit the number of returned tokens to 2 in order to support values that contains '='.
I will send a PR over the weekend.
tx

CumpsD:
Thanks

PR #94, opened 2017-02-03
Don't restrict to Information level when no mum is specified

### LINKED ISSUE #93: Appsettings; Minimum level from Serilog is not being respected
This sink defaults to a minimum level of Information if it is not overwritten specifically. This only occurs when initialising from AppSettings.

The code pasted below will allow you to reproduce the issue.

```
<appSettings>
  <add key="serilog:minimum-level" value="Debug" />
  <add key="serilog:using:RollingFile" value="Serilog.Sinks.RollingFile" />
  <add key="serilog:write-to:RollingFile.buffered" value="false" />
  <add key="serilog:write-to:RollingFile.pathFormat" value="C:\Logs\SerilogTest\Log-{Date}.txt" />
  <add key="serilog:write-to:RollingFile.retainedFileCountLimit" value="30" />

  <add key="serilog:using" value="Serilog.Sinks.Elasticsearch" />
  <add key="serilog:write-to:Elasticsearch.nodeUris" value="http://localhost:8888" />
  <add key="serilog:write-to:Elasticsearch.indexFormat" value="test-{0:yyyy.MM}" />
  <add key="serilog:write-to:Elasticsearch.templateName" value="test-template" />
  <add key="serilog:write-to:Elasticsearch.batchPostingLimit" value="1" />
  <!--<add key="serilog:write-to:Elasticsearch.minimumLogEventLevel" value="Debug"/>-->
</appSettings>
```

```
Log.Logger = new LoggerConfiguration()
    .ReadFrom.AppSettings()
    //.WriteTo.Elasticsearch(new ElasticsearchSinkOptions(new Uri("http://localhost:8888"))
    //{
    //    AutoRegisterTemplate = true,
    //    TemplateName = "test-template",
    //    IndexFormat = "test-{0:yyyy.MM}",
    //    BatchPostingLimit = 1
    //})
    .CreateLogger();

var file = File.CreateText("C:\\Logs\\SeriLogErrors.txt");
Serilog.Debugging.SelfLog.Enable(TextWriter.Synchronized(file));

Log.Logger.Debug("I am a debug message");
```

nblumhardt:
Sinks don't normally restrict the minimum log event level unless explicitly told to do so. The configuration overload accepting `ElasticsearchSinkOptions` respects this convention, but the appsettings compatible overload does not.

The change removes this default limitation. This might be surprising to some users - if we want to go so far as to consider it a major-version incrementing breaking change, we might wish to also rename the parameter `restrictedToMinimumLevel`, which is the name used by all other Serilog sinks for this (ES is alone in calling this parameter `minimumLogEventLevel`).

Thoughts?

Fixes #93.

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
+++ b/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
@@ -49,10 +49,6 @@ namespace Serilog
             this LoggerSinkConfiguration loggerSinkConfiguration,
             ElasticsearchSinkOptions options = null)
         {
-            //TODO make sure we do not kill appdata injection
-            //TODO handle bulk errors and write to self log, what does logstash do in this case?
-            //TODO NEST trace logging ID's to corrolate requests to eachother
-
             options = options ?? new ElasticsearchSinkOptions(new[] { new Uri(DefaultNodeUri) });
 
             var sink = string.IsNullOrWhiteSpace(options.BufferBaseFilename)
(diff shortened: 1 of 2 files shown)


mivano:
It is indeed a frequent question why the sink does not respect the filtering. As the behavior is different anyway inside the sink, I think it would be better to merge and directly change the name. That would at least trigger something so people are aware :-)

PR #97, opened 2017-02-17
support config global headers that contains equal sign ('=') in its v…

AmirSasson:
support config global headers that contains equal sign ('=') in its value
suggested by @CumpsD, on PR #87


mivano:
Good addition, thanks

PR #107, opened 2017-05-17
Added option to not store rendered message

ipinak:


mivano:
What is your use case that you want to disable this option? You can also implement this now by using your custom formatter instead?

ipinak:
I've used this 

```
"WriteTo": [
      {
        "Name": "ColoredConsole",
        "Args": {
          "outputTemplate": "{Timestamp:yyyy-MM-dd HH:mm:ss.fff} [{Level}] {Message}{NewLine}{Exception}"
        }
      }
    ]
```
But it seems that it logs `message` twice.

mivano:
What you show here is the output template for a colored console sink, not the ES one. 

If you want to change the output, use a customer formatter. Your change here is a breaking one as it impact signature even while it is optional.

mivano:
Still missing the reason why you want this change? Closing for now.

ipinak:
Sorry, for not replying. Close it.

PR #109, opened 2017-05-24
CSPROJ + ES.NET 5

ffMathy:
- Upgraded to CSPROJ format (this was done automatically by Visual Studio 2017 when loading the solution). Confirmed to be building in Visual Studio 2017 Community with a clean repository.
- Upgraded NuGet packages of the project to latest versions, including Elasticsearch.Net 5.


ffMathy:
Hmmm build server is really not letting me build this. Builds fine locally. I think it's because the build server uses `1.0.4` of the .NET CLI, and not `1.1.0`.

Do you know why? Like I said, everything works locally, even the unit tests.

ffMathy:
I think you might need to configure the AppVeyor build settings to use Visual Studio 2017 for building, or at least a later version of `dotnet.exe`. Can you do that?

ffMathy:
@mivano any chance you can take a look at this? My team is kind of stuck until we get this fixed.

mivano:
Sorry I have been away some days. I will try to look into it today!

mivano:
In the appveyor.yml file, can you add this line and try again?

```yml
image: Visual Studio 2017
```

ffMathy:
Thank you so much for the attention - it is really important for us, and your sink is already a central part of our daily work.

It still doesn't build though. I think the image won't work, it has something to do with the version of `dotnet.exe` I believe.

mivano:
So it either did not pick up the right image (I cant see which one it selects, but the default one is VS 2017, although it should respect the config file first) or it is another kind of error. 
Can you output the build/dotnet version into the console?

ffMathy:
Proof of point. According to the build log, the following targets file is used:
`C:\Program Files\dotnet\sdk\1.0.4\Sdks\Microsoft.NET.Sdk\build\Microsoft.NET.Sdk.targets`.

Note the version `1.0.4`.

mivano:
hmm, looks like they did include the latest VS2017 recently (https://www.appveyor.com/updates/).

ffMathy:
In that very same link, it says ".NET Core SDK 1.0.4". We need `1.1.0`.

.NET Core is now standalone from Visual Studio.

ffMathy:
Alright, it seems that version `1.1.0` of the .NET Core SDK still gives `1.0.4` of the `dotnet.exe` file - my bad.

I am now investigating fixes suggested here: https://github.com/Microsoft/vsts-tasks/issues/3762

mivano:
I cannot find a way to force it to use another dotnet version. I can only select another environment (like VS 2017), but the same you can do in the yml file.

ffMathy:
Is there any chance you could try to help fiddling with the `.csproj` files with me to get it working?

I'll try creating a new .NET Core project tonight to see how that looks, and try to figure out what is wrong.

ffMathy:
Build now "completes" by ignoring errors, but gives a NuGet restore issue. Do you have any idea of how to prevent it? See the build log.

ffMathy:
I don't even understand why it tries to install `JetBrains.Annotations`. I don't see it in any of the `.csproj` files anywhere. Do you know how this dependency is somehow added?

And what about the other one as well, `YamlDotNet`? Where does this come from, and why is it needed?

ffMathy:
Aha! Finally found the issue.

It's because I bumped to .NET Core 1.1, which is no longer supported by `gitversion`. See the issue https://github.com/ah-/dotnet-gitversion/issues/4.

ffMathy:
Alright, this is now ready for review!

ffMathy:
@mivano care to take one (hopefully) final look?

mivano:
Ah glad you found the underlying issue. However gitversion is doing the versioning, so how to fix that now? We can use the standalone tool but it also looks like it has support for appveyor (https://gitversion.readthedocs.io/en/stable/build-server-support/build-server/appveyor/)

ffMathy:
Alright @mivano, it now uses the command line tools. However, if I download the artifacts from the build server (the `.nupkg` file), rename it to a `.zip` file, extract it and look at the libraries attached, their assembly versions are still `5.1.0`.

Is this still how it's supposed to work? According to the build log, GitVersion did actually run successfully and did find assembly files to patch.

The package has version `5.2.0` though which I specified.

mivano:
The dotnet gitflow library updated the version number inside the project json file. So that trick wont work anyway anymore. So somehow we need to inject the current version number in the assembly again.

ffMathy:
But it actually does this! Currently it even appends the pull request to the assembly file version too. 

I think the reason 5.1 is chosen is simply because there has been no tag or merge of 5.2 yet. Isn't this how it normally behaves @mivano?

ffMathy:
What I say is that we can try this and see if it goes well. Worst case scenario would be that there would still be a 5.2 version of the package, but the DLLs would read 5.1.

If that's the case, we can fix it quickly and then publish a new version. How do you usually test that this works? Should I turn this into a release branch somehow? 

That could potentially work, if you just merged this into a release branch called 5.2.

But my knowledge of gitversion is limited.

mivano:
Yes, it was conditional. AppVeyor checked out the PR in some way that made the gitversion not work. I merged all now to dev. If that build, I merge to master for a package.

mivano:
In master and deployed to nuget. Thanks for the PR!

Issue #102, opened 2017-03-08
Missing {"@ in body


JakobVork:
Hi,

I have been using serilog for some time now, and have noticed that we once in a while are missing some logs. When I looked into what might caused this I found the following:

First I created a test API that recieved a call and when logged it into a bufferFile (That never miss a log btw). 
I when created a program, that sent 5000 requests to the API. I could see 5000 logs in the bufferFile, but only 4999 logs when I looked at kibana.

I when looked at the network traffic, and found the following:
![image](https://cloud.githubusercontent.com/assets/10922674/23705814/d0c69882-040b-11e7-9f4b-113ce3fef522.png)

If we take a look on the server running elasticsearch it gives this error:
![image](https://cloud.githubusercontent.com/assets/10922674/23705689/58c05756-040b-11e7-97d8-5d2a3778206b.png)

So, from what I can see is that serilog sometimes cuts {"@ away from timestamp. I've tried to run it multiple times, but it does not always happen. (Often need to run it 4-10 times before I see the error again). It is however always the second log that fails.
The bufferFile does not cut away {"@ as shown below:
![image](https://cloud.githubusercontent.com/assets/10922674/23705542/98e6156a-040a-11e7-8019-56763d742308.png)

Not sure why this happens, and especially why this only happens sometimes?

Currently running with ElasticSearch 5.2.1 and Kibana 5.2.1.
The nuget package I use is Serilog.Sinks.Elasticsearch version 5.0.0 (depend on serilog version 2.2.1)

nblumhardt:
This sounds like https://github.com/serilog/serilog-sinks-seq/issues/28 - caused by `HttpLogShipper` expecting the (three-byte) Unicode BOM, but more recent versions of Serilog.Sinks.File (and RollingFile) defaulting to no-BOM.

JakobVork:
I have now tried to upgrade Serilog.Sinks.File to version 3.2.0, but the behavior is still the same. (Even though, the original dep. was => 3.0.1, and therefore still newer than the mentioned in the link you provided).

nblumhardt:
If the issue is the linked one, a fix will be required in this repository.

The fix is: https://github.com/serilog/serilog-sinks-seq/commit/6c0b308efab7aba2062f00b1b83d3749e6d43bfb

If anyone can PR this I can help with review etc.

Cheers,
Nick

mivano:
This one is merged in v 5.5

Issue #25, opened 2016-02-21
ElasticsearchLogShipper has the wrong success criteria for Bulk operations


nblumhardt:
Moved from: https://github.com/serilog/serilog/issues/599

> When reading from a durable file and uploading bulk actions to the elastic search server, the log shipper checks the HTTP response for a Success Code and if true it writes the bookmarks and moves on.

megakid:
Due to our monotonic bookmarking in the durable shipper, as noted in the linked issue, it is difficult to know what to do when a single document in the middle of a `bulk index` fails.

We can use the `_id` field to stop us inserting duplicates here.  If we generate a unique `_id` when we save the files to the durable file, we can use this in each of our `index` commands to allow ElasticSearch to either: reject duplicate `_id`s that are already indexed using the `create` operation or "upsert" (updating any existing document with the same `_id` or create if nonexistent) using the `index` operation.  Only if the entire batch passes do we bookmark and continue.  Perhaps we could retry a batch a certain number of times before bookmarking it anyway?

Lanoli:
What's the status on this issue ? 
@konste stated on https://github.com/serilog/serilog/issues/599 that a PR was made with NEST 2.0, but I can't seem te find it.
@mivano was working on a feature branch  https://github.com/serilog/serilog-sinks-elasticsearch/tree/handlefailures , but this was never finished or merged to dev. 

It would be nice to just parse the bulk response and log any failures. Retries are less important to me, as most failures will most likely be the result of malformed requests.

Are there still any plans to continute this work, or will it be discarded?

konste:
This is custom modification of the official Serilog ES Sink, done for two reasons:
1. To be able to apply modified ES Json Formatter;  
2. To modify EmitBatch method so that it records possible server side failures to internal log;

[ElasticsearchSinkMod.cs.txt](https://github.com/serilog/serilog-sinks-elasticsearch/files/1039498/ElasticsearchSinkMod.cs.txt)

Lanoli:
Thanks @konste . I'm using the DurableElasticsearchSink though, the parsing of the response happens in the ElastsearchLogShipper. Which means I'ld have to copy a lot more code.  I'll first try fixing the problem in my fork. I see you are using Nest to parse the response to BulkResponse. Something I already tried and which wasn't 100% corrent. So I'll look into that. Problem is the overlap with @mivano 's work, of which I wonder why it was discontinued.

Lanoli:
#110

[...]

PR #110, opened 2017-05-30
Missing {"@ in body #102. ElasticsearchLogShipper has the wrong success criteria for Bulk operations #25

Lanoli:
Enable reading from UTF-8 files with or without BOM.


Lanoli:
Deserialize bulkresponse & log failed operations

mivano:
You introduce the full NEST library here. We try to stick with the more lightweight Elasticsearch.net library only. Can we do without NEST?

Lanoli:
NEST contains the BulkResponse class which uses BulkResponseItemJsonConverter to deserialize.

mivano:
That is true, but it also introduces a lot of other stuff. You can use a dynamic or parse it yourself without introducing NEST.

Lanoli:
I mailndy used NEST because I don't know the responsetypes of all the different elasticsearch versions. This way you get a battletested, maintained deserialisation.  Copying these classes over to this project is an option to make it more lighteweight.

mivano:
See here (https://github.com/serilog/serilog-sinks-elasticsearch/blob/handlefailures/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs#L60) for a way to parse without the NEST dependency. We would like to stick with the low level elasticsearch.net library as much as possible.

Lanoli:
Why not copy the Bulkresponse and jsonconverter from NEST over to this project ? Those are well tested and used by the community.

Lanoli:
I removed NEST and changed the parsing to use dynamics + merge with DEV.

lrolsen:
@Lanoli  Are there any plans to get these changes into master?

Lanoli:
@mivano ,

olegKoshmeliuk:
@mivano any chance to have such feature merged into master?
I have few point how to improve this PR.
1. remove custom created id in at sink, since items array order is same as request items (https://www.elastic.co/guide/en/elasticsearch/guide/current/bulk.html)
2. Add same code for not durable sink
3. expose error event to user code via callback or event (in ElasticsearchSinkOptions). This will helpul if you have mapping errors ( like "failed to parse [fields.LogId]" expected string got object). so users can log them back into elasticsearch.

If you don't mind i can create pull requests with my changes.

mivano:
I merged your changes and dev into this branch: https://github.com/serilog/serilog-sinks-elasticsearch/tree/Lanoli-dev
I also added some other improvements that should help with error detection and recovery. I m not completely satisfied with it yet, so please have a look. 

I also added a sample project and a docker compose file to set up an environment.

Regarding your points:
 1. Yes I think that is not needed, see also my implementation in https://github.com/serilog/serilog-sinks-elasticsearch/blob/0c9bdd35e2aa475b4c84ca1cd7d4286c4c6be76a/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs#L63
 2. Although I like the ability for the durable sink, it adds duplication which should not be there.
 3. I added this option: https://github.com/serilog/serilog-sinks-elasticsearch/blob/0c9bdd35e2aa475b4c84ca1cd7d4286c4c6be76a/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkOptions.cs#L236-L252 which allows you to act on issues.

So please have a look and give your thoughts, thanks!

olegKoshmeliuk:
You meed to update durable sink to respect EmitEventFailure option.
For example from your test applcation
image
it this case FailureSink will be not called and it's not clear for sink clients. 
For durable sink you can log into failure sink logevent with payload and error.
Rest is good. 👍

mivano:
Thanks for the review. I will make the changes and do a PR.

mivano:
See #127

PR #124, opened 2017-09-27
Added support for Pipeline in Elasticsearch sink options

ajviradia:
Adding support for pipeline to enable some wrangling and enrichment, courtesy Elasticsearch!


mivano:
Thanks! Great addition. I merged to dev and master but fighting now with pushing the packages to nuget server.

Issue #138, opened 2017-12-20
No `LoggingLevelSwitch` option in the extension method


piaste:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [x] New Feature

**Please describe the current behavior?**

The `LoggerSinkConfiguration.Elasticsearch()` method allows you to specify a minimum logging level, but does not have an option to pass in a `LoggingLevelSwitch` object through which the minimum logging level can be changed at runtime.

Adding this object would allow a user to set a different log level for the ElasticSearch sink compared to any other sinks in use.

Note that the object can simply be passed immediately to the `LoggerSinkConfiguration.Sink()` method and does not require any actual implementation.

**Please describe the expected behavior?**

There should be a `LoggingLevelSwitch` property in the `ElasticsearchSinkOptions` class to enable use of this feature. Many Serilog sinks (including Console and File) have this option.

I will shortly submit a PR to add this property.

mivano:
Merged and released.

PR #139, opened 2017-12-20
Added logLevelSwitch option

piaste:
Allows the sink to be passed a `LoggingLevelSwitch` object as an alternative to a plain minimum log level. See #138.

No. The switch is the last argument in the public method in which it appears.

I don't think unit tests would be appropriate here since all the code is in Serilog itself, the PR simply passes in an extra argument. 

I've tested it manually and the switches operate correctly as expected (open two ES sinks, pass different switches, set different minimum levels, log only goes to one).

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
+++ b/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
@@ -58,7 +58,7 @@ namespace Serilog
             return loggerSinkConfiguration.Sink(
                 sink,
                 restrictedToMinimumLevel : options.MinimumLogEventLevel ?? LevelAlias.Minimum,
-                levelSwitch : options.LoggingLevelSwitch
+                levelSwitch : options.LevelSwitch
             );
         }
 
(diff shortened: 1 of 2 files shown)


src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
```diff
@@ -89,7 +94,8 @@ public static class LoggerConfigurationElasticsearchExtensions
             string bufferBaseFilename = null,
             long? bufferFileSizeLimitBytes = null,
             long bufferLogShippingInterval = 5000,
-            string connectionGlobalHeaders = null)
+            string connectionGlobalHeaders = null,
+            LoggingLevelSwitch loggingLevelSwitch = null)
```
nblumhardt: In the other public APIs exposed by Serilog, [we call this `levelSwitch`](https://github.com/serilog/serilog/blob/dev/src/Serilog/Configuration/LoggerSinkConfiguration.cs#L71) - would be good  to be consistent (if this change is accepted).

piaste: Good point. Will align.

mivano:
Thanks for the PR, nice addition.

Issue #129, opened 2017-11-10
Auto-registered template doesn't support ElasticSearch 5 string types.


plmaheu:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [x] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
5.4

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [ ] netCore 2.0
- [ ] netCore 1.0
- [x] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**
The provided dynamic mapping for strings doesn't work for ElasticSearch 5.x. The raw field isn't being mapped at all.

It's using string type, as in
```
 "string_fields": {
              "match": "*",
              "match_mapping_type": "string",
              "mapping": {
                "fields": {
                  "raw": {
                    "ignore_above": 256,
                    "index": "not_analyzed",
                    "type": "string"
                  }
                },
                "index": "analyzed",
                "omit_norms": true,
                "type": "string"
              }
```
**Please describe the expected behavior?**
Should use 'text' and 'keyword', as per 5.x syntax.

mivano:
Thanks for reporting. Something we need to fix. Care for a PR?

quetzalcoatl:
I just tried to make it work with ES 6.0, and there's more to it than just the string->text replacement. It seems that 'string' should stay at least in `match_mapping_type`, I got some errors when I changed it there.

Furthermore, all `index` are now boolean, so "analyzed/not_analyzed" have to be changed to true/false respectively.

Next thing, one I lack knowledge to properly fix, is `_all` - it was deprecated in 5.x, and now in 6.x it has been removed. I found information that it is possible to emulate it very well using dynamic templates that match all fields and making them `copy_to` their contents to some other new field (with custom name and index:true,store:false) that now can be used in lieu of non-existent `_all` - however I failed trying to configure that.

Last thing I noticed is .. that in 6.x the `_default_` has been deprecated. It will work for some time and then evaporate like _all did. The rationale is that now there's a single type per index, so _default_ has no sense. OK. But I'm totally new to ES so I have no idea how to fix that here in the mapping :)

[...]

PR #142, opened 2018-01-05
Get rid of 'deprecated' warnings in Elastic 5.6.2

VanHuychemG:
Our Elastic 5.6.2 logs are flooded with below errors:
The [string] field is deprecated, please use [text] or [keyword]
[omit_norms] is deprecated, please use [norms] instead with the opposite boolean value

Updated the schema following https://www.elastic.co/blog/strings-are-dead-long-live-strings


CumpsD:
Looking forward to this, we are seeing the same

CumpsD:
When reading the issues, it seems in ES 6 the deprecation turned into removal, so this PR might fix ES 6 support too :)

quetzalcoatl:
LGTM - but I'm no expert here :) Changes are similar to what I did to patch up the same set of issues for me (see code I provided in comments in issue #129). However, I didn't know about `keyword` type (sadly, I didn't stumble on that strings-are-dead article) so I had to comment out 'ignore_above' as well. Good to know! This piece is obviously much better than mine.

One thing I noticed is that this is just a change to mappings. Since newest ElasticSearch changed the schema (like 'index' is now true/false), then IMHO applying this will break support for lower versions of ElasticSearch. That may be OK, but then, [at least one person was concerned some time ago](https://github.com/serilog/serilog-sinks-elasticsearch/issues/129#issuecomment-351444263)

VanHuychemG:
I'll have a go at creating some kind of template provider, that can be set from the options ,something like pre-version5, version5, version6.

CumpsD:
Awesome! Looking forward to get this merged! // @mivano

mivano:
Thanks for the all the work. I will have a look soon.
Op do 11 jan. 2018 om 13:44 schreef David Cumps <notifications@github.com>

mivano:
Great addition and thanks for the backward capability option!

CumpsD:
Awesome!

Issue #143, opened 2018-01-18
Configure ExceptionAsObjectJsonFormatter via AppSettings


iss0:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [x] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
latest

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [ ] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [x] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**
I am not able to configure the ExceptionAsObjectJsonFormatter  via App.Config or Web.Config.

**Please describe the expected behavior?**
I would like to configure it that way. :-)

mivano:
Old issue, cleaning up

PR #145, opened 2018-01-18
Configure ExceptionAsObjectJsonFormatter via configuration

iss0:
#143

No

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
+++ b/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
@@ -12,15 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
 using Serilog.Configuration;
 using Serilog.Core;
 using Serilog.Events;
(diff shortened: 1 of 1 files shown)


mivano:
Thanks for the PR. Adding parameters, even optional, is breaking the signature. Not a big deal, but something to be aware of. I see what you want to do, but this makes it very specific. Could you not create your own method extension in your project with that specific code, name it ElasticsearchCustom and use that instead for your configuration needs? Now it is linked to this specific formatter using magic strings.

iss0:
You're quiet right about the magic string stuff. It is really not a good way to do it. At least I could've used nameof(ExceptionAsObjectJsonFormatter) in the case.

In general using an extension method in my application to configure the custom formatter means I can not use App.Config or Web.Config to configure the sink at all or can I?

mivano:
I believe the code that populates the config looks for matching method names. So use ElasticsearchCustom in this case as the name of the sink. Not tested this, but might be a path to explore?

iss0:
Oh yes! I did not know that. Sounds like the perfect solution for me. I will have a look if this works for me. Thank you!

iss0:
Just a short heads up. It did actually work quite well. I am able to configure it via my own extension method. Thinks this can be closed for good.

mivano:
Perfect, thanks for the follow up.

Issue #149, opened 2018-01-23
Configure more ElasticsearchSinkOptions through AppSettings


letargis:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [X] New Feature

**Please describe the current behavior?**
Only limited subset of ElasticsearchSinkOptions properties allowed to configure through AppSettings.

**Please describe the expected behavior?**
I understand that it is hard to configure properties of Func<>, but I propose to extend number of simple parameters. For my current task I need to manage numberOfShards and numberOfReplicas properties from configuration, so I propose to add to appSettings following properties:

- autoRegisterTemplate
- autoRegisterTemplateVersion
- overwriteTemplate
- numberOfShards
- numberOfReplicas

mivano:
See my feedback at the PR.

PR #150, opened 2018-01-24
Extend LoggerConfiguration.Elasticsearch with additional attributes

letargis:
#149

No

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
+++ b/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
@@ -21,6 +21,7 @@ using Serilog.Core;
 using Serilog.Events;
 using Serilog.Sinks.Elasticsearch;
 using System.Collections.Specialized;
+using System.ComponentModel;
 
 namespace Serilog
 {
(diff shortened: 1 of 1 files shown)


mivano:
Thanks. Adding parameters, even optional, is a breaking change. And although I see why you want this, I m a bit afraid we are making this method longer and longer while we actually introduced an object to avoid this.
@nblumhardt any tips how to handle this? I see more request for actually wanting to manage the config of the sink in the appSetting and adding the options to the method seems to be the only way.

letargis:
So, @mivano, @nblumhardt  what will be the conclusion?
For me it looks, that LoggerConfiguration.Elasticsearch method with multiple parameters has the only purpose to support appSettings configuration. So it is natural to extend it with all parameters, which could be configured via config file.
Otherwise, I should use LoggerConfiguration.Elasticsearch with object parameter and manage settings myself. This is strange, because Elasticsearch sink supports appConfig settings.

mivano:
@nblumhardt any tips/suggestions for this?

nblumhardt:
Howdy! In the past, where we've had to do this, the process has been:

1. Create a new overload with all of the existing parameters (excluding any obsolete ones) plus the new parameters
2. Make the old method forward to the new one (i.e. implement it by calling new method)
3. Since this still won't play nicely with overload resolution:
   - Remove all of the default values from the old method
   - Mark the old method as `[EditorBrowsable(EditorBrowsableState.Never)]`
4. If there's no valid scenario in which every single parameter on the old method could be validly supplied (i.e. some are mutually-exclusive), `[Obsolete]` the old method

This should avoid all source and binary breaking changes. It's at the cost of keeping a method with the old signature around, but it's proven to be unobtrusive so far.

See, for example: https://github.com/serilog/serilog-sinks-file/blob/dev/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs#L59

HTH, let me know if you need more details.

letargis:
I've created new method and modified old one as suggested by @nblumhardt . Moreover, I've added some additional parameters which could be configured via config file.

mivano:
Sorry, it took a while but merged now. Thanks for the PR!

PR #157, opened 2018-03-01
Explain use of `AutoRegisterTemplateVersion` in the README

piaste:
It took me a while to figure out that the reason the sink failed to start with a somewhat cryptic "Failed to execute" error was that, while I had updated all libraries to ES6-compatibile versions, I had not set the  `AutoRegisterTemplateVersion` option and so the template sent was not ES6-compatible (it used "string").

This PR explains this snag in the readme.

Note: for a future PR, would it make sense to deprecate the manual option and automatically detect the ES version when attempting to register a template? It can be fetched by a GET request at the root URL, which returns:

```    {
      "name" : "8hV7EYy",
      "cluster_name" : "docker-cluster",
      "cluster_uuid" : "bC1bTjVuTiyWoaprIIZIIw",
      "version" : {
        "number" : "6.2.1",
        "build_hash" : "7299dc3",
        "build_date" : "2018-02-07T19:34:26.990113Z",
        "build_snapshot" : false,
        "lucene_version" : "7.2.1",
        "minimum_wire_compatibility_version" : "5.6.0",
        "minimum_index_compatibility_version" : "5.0.0"
      },
      "tagline" : "You Know, for Search"
    }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Register the sink in code or using the appSettings reader (from v2.0.42+) as sho
 var loggerConfig = new LoggerConfiguration()
     .WriteTo.Elasticsearch(new ElasticsearchSinkOptions(new Uri("http://localhost:9200") ){
              AutoRegisterTemplate = true,
+             AutoRegisterTemplateVersion = AutoRegisterTemplateVersion.ESv6
      });
 ```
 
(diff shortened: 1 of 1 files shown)


mivano:
Thanks! We opted, for now, to use the v2 to not break existing installs, but a detection would even be better.

Issue #203, opened 2018-12-11
log elastic mapping errors as errors in elasticsearch


pierresetteskog:
**A few questions before you begin:**

> Is this an issue related to the [Serilog core project](https://github.com/serilog/serilog) or one of the [sinks](https://github.com/serilog/serilog/wiki/Provided-Sinks) or 
[community projects](https://github.com/serilog/serilog/wiki/Community-Projects).  
This issue list is intended for Serilog Elasticsearch Sink issues. If this issue relates to another sink or to the code project, 
please log on the related repository.  Please use [Gitter chat](https://gitter.im/serilog/serilog) and [Stack Overflow](http://stackoverflow.com/questions/tagged/serilog)  for discussions and questons.

**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [x] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [x] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**
Its easy to by accident log same field with different datatypes and get errors like:

Received failed ElasticSearch shipping result 400: {"type":"mapper_parsing_exception","reason":"failed to parse field 

And its easy to miss errors in  selflog
**Please describe the expected behavior?**
It would be nice if all not connection error problem was logged as errors in elasticsearch as string.
Or at least the excption is also sent to   Action<LogEvent,Exception> ElasticsearchConfiguration.FailureCallback

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**

mivano:
I do see what you mean, due to the batching and background work, it is difficult to spot issues. it is also not an audit sink, so exceptions are hidden. The use of the `EmitEventFailureHandling` option might be useful here?

mivano:
Old issue, cleaning up. Still facing this issue, feel free to reopen.

Issue #194, opened 2018-10-02
Unexpected behaviour while using IndexDecider


simonare:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [X] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
Serilog.Sinks.Elasticsearch Version 6.5.0

**What is the target framework and operating system? See
- [X] netCore 2.1.4
- [ ] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**
 I am trying to use IndexDecider for storing RequestLogs to different Index on ElasticSearch. However, the code does not work. Basicly, It creates the new Index but does not populate in it. Instead, sink sends logs to default Index which is logstash-{0:yyyy.MM.dd}

**Please describe the expected behavior?**
The expected behavior would be to send logs into the request-log-{0:yyyy.MM.dd}

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**

The configuration and code is as below:

```
  var loggerConfiguration =  new LoggerConfiguration()
                .ReadFrom.Configuration(_configuration.GetSection("Logging"))
                .WriteTo.Elasticsearch(new ElasticsearchSinkOptions(new Uri("http://cwlog.somehost.com:9200"))
                { 
                    IndexDecider = (e, offset) =>
                    {
                        if (e.MessageTemplate.Text.Equals("<{EventID:l}> Http Request Method: {Method} Url: {Url}"))
                        {
                            var indexName = $"request-log-{offset:yyyy.MM.dd}";
                            return indexName;
                        }

                        return $"logstash-{offset:yyyy.MM.dd}";
                    },
                    CustomFormatter = new ExceptionAsObjectJsonFormatter(true)
                });
```

```
{
          "Name": "Elasticsearch",
          "Args": {
            "nodeUris": "http://cwlog.creditwestbank.com:9200",
            //"minimumLogEventLevel": "Debug",
            "autoRegisterTemplate": true,
            "autoRegisterTemplateVersion": "ESv6",
            //"customFormatter": "Serilog.Sinks.Elasticsearch.ExceptionAsObjectJsonFormatter, Serilog",
            "bufferBaseFilename": "c:\\logs\\dev\\westapi\\es-buffer",
[...]

Issue #179, opened 2018-06-14
Port buffer file improvements from Serilog.Sinks.Seq


nvivo:
Currently, this sink is using RollingFile as buffer, and it has some serious limitations, that whether the buffer file grows without limits, or the file is size-limited and the sink stops working after that. This is making the durable mode unusable with this sink.

I'm currently using the sink without durable mode, but this is a problem as some events may go missing for various reasons. The Seq sink had the same issues, and they [were fixed on version 4.0](https://github.com/serilog/serilog-sinks-seq/releases/tag/v4.0.0). It would be great if these changes, that is https://github.com/serilog/serilog-sinks-seq/pull/92 and https://github.com/serilog/serilog-sinks-seq/pull/99 would be ported to this sink.

nvivo:
Is it possible to pay to get this implemented? Is anyone interested in resolving this?

mivano:
Would love to have a better/improved version of the buffer, unfortunately too little time to implement. IMHO it would also be nice if this was a more generic Serilog buffer which would pick up events when the configured sink was unable to work.

nvivo:
I currently have this sink and the seq sink logging the same events. For some reason, I have lots of missing events in elastic search that are not logged nor reported in the failure sink/self log. I enabled the selflog, and eventually I get a few failures, but that is probably 1% of what is missing on elastic search.

I'm betting the lack of durable mode is the problem, but that's just a guess. I'll try to debug this at some point.

mivano:
Old issue, cleaning up. There have been some improvement. Still facing this issue, feel free to reopen.

Issue #166, opened 2018-04-16
Skip sending oversized data to ES


MinhNguyenDev:
I had a same issue as #112, one line in the log file was too big and when the ElasticsearchLogShipper tried to send the data, it failed and kept retry over and over again, which could be costly if using AWS Elasticsearch. Could we make a change to skip the large size record ? It will fail anyway and make the whole log shipper stop working and no way to recover unless we delete the line manually.
Here is some sample code:

```
                        while (count < _batchPostingLimit && TryReadLine(current, ref nextLineBeginsAtOffset, out nextLine))

                        {
                            if (nextLine.Length > 1000000) // We can make this number configurable
                            {
                                SelfLog.WriteLine("Line is too long. Length {0}. Skip sending to ElasticSearch", nextLine.Length);
                                continue;
                            }
                            // Add payload
                            ++count;
                        }
```

Thanks,

mivano:
Sounds good. Care for a PR? Make it configurable would be good with a default that does not check for this?

mivano:
Merged, closing issue

PR #205, opened 2018-12-18
Using durable seq

pierresetteskog:
#179 Port buffer file improvements from Serilog.Sinks.Seq 
#203 is fixed with BufferCleanPayload example can be found in Serilog.Sinks.Elasticsearch.Sample\Program.cs
#194 with BufferIndexDecider you can see the logrow and it is runned for every row see example program
#204 BufferFileCountLimit see example program
#166 SingleEventSizePostingLimit see example program
#146 fixed by using code from sink.seq

No not what i know

almost no logical changes is made on the code just structural refactoring..
**Other information**:
code is mostly from sinks.Seq and placed in the folder Durable some classes like LogShipper PayloadReader is made generic with no dependencies to seq. Would be nice to place in some generic nuget package to share between different sinks.
Lots of settings was just working in file mode or in memory mode. 
Maybe two different configure classes would be more clear.
In the example program Program.SetupLoggerWithPersistantStorage
I have tried to include all settings which is available in the persistant mode.
Sorry for the big pull request but it was a big task to move over persistant code from sink.seq.


src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/Durable/ElasticSearch/ElasticSearchLogShipper.cs
```diff
@@ -0,0 +1,71 @@
...
+using Serilog.Core;
+using Serilog.Debugging;
+using Serilog.Sinks.Elasticsearch.Durable;
+
+namespace Serilog.Sinks.Elasticsearch.Durable
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class ElasticSearchLogShipper : LogShipper<List<string>>
```
mivano:
```suggestion
    public class ElasticsearchLogShipper : LogShipper<List<string>>
```
Convention is to use Elasticsearch instead of ElasticSearch

src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/Durable/ElasticSearch/ElasticSearchPayloadReader.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Serilog.Sinks.Elasticsearch.Durable
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class ElasticSearchPayloadReader: APayloadReader<List<string>>
```
mivano:
```suggestion
    public class ElasticsearchPayloadReader: APayloadReader<List<string>>
```

mivano:
Thanks for your PR! Looks good. I m not actively using this sink at the moment, so it is a bit hard for me to see if this will work out fine. Adding the Seq implementation is certainly a nice go forward and the other changes look sensible. 
There are however some breaking changes; types are changed and method signature has some added, although default null, parameters. Not a big issues, but would be good to bump the major. 

Can you also alter the `changes.md` and the `readme.md` with the relevant details? Then we can merge to dev and roll out a pre release package first.

thanks!

pierresetteskog:
Greate i will be on a vacation a week now will do it after that!

pierresetteskog:
I didnt know which version to use but you mentioned major version bump so i wrote 7. Please let me know if I should do something more. :)

mivano:
Package: https://www.nuget.org/packages/Serilog.Sinks.ElasticSearch/6.5.0-alpha0045

PR #211, opened 2019-02-05
fix title of changelog

tomap:
small typo in markdown

Will probably break the internet


mivano:
Nice catch. Tx for the PR!

PR #222, opened 2019-03-13
add .netstandard2.0 target

### LINKED ISSUE #196: when logging an exception the "ClassName" property remains empty
**Does this issue relate to a new *feature* or an existing *bug*?**
- [x] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
Serilog.Sinks.ElasticSearch, version 6.5.0

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [x] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**

When an exception is logged and sent to Elasticsearch, in Kibana I can see the `exceptions` property. Several subproperties in `exceptions` are filled as expected, like `Message`, `Source`, `StackTraceString`, but the property `ClassName` remains empty. This appears to be related to the fact that `NO_SERIALIZATION` is defined when compiling for netstandard1.3. (Because if `NO_SERIALIZATION` is defined, then `var className = string.Empty;` is used.)

**Please describe the expected behavior?**

I expect the `ClassName` property under the `exceptions` property to not be empty, but instead contain the full name of the type of the exception (e.g. `System.InvalidOperationException`).

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**

I created a fork and I think I fixed it there (see the last 2 commits (633ae5efc19c8e5b2b64c4de192a057c88d7de9e and 338587206a101ba4d5732e03330a129a6da3f3d1) on https://github.com/roelandvh/serilog-sinks-elasticsearch) using `exception.GetType().FullName`, but I'm not sure that is a valid solution for this case (perhaps `.GetType` is too expensive in this context?).

Also I had to tweak the .csproj file to force `NO_SERIALIZATION` to be set (which I did not commit) because apparently everything was building and running as if targetting `net45`. (When I did this, the test actually failed when I removed my fix, and succeeded after re-applying the fix, so I'm confident it works as expected when `NO_SERIALIZATION` is set.)

### LINKED ISSUE #210: Add .Net Standard 2.0 target
According to [Open-source library guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting) it would be good to target .Net Standard 2.0 to benefit from having a smaller package graph.

I would appreciate to help with it.

vhatsura:
to fix #210 and fix #196  

A few comments about changes:

* `DOTNETCORE` constant is used only in tests projects. By this reason it was removed from `Serilog.Formatting.Elasticsearch` and `Serilog.Sinks.Elasticsearch` projects
* `NO_SERIALIZATION` and `NO_TIMER` constants are not used in `Serilog.Sinks.Elasticsearch` project and was removed from `netstandard1.3` target framework
* [System.Exception.GetObjectData](https://docs.microsoft.com/en-us/dotnet/api/system.exception.getobjectdata?view=netframework-4.7.2#applies-to) is accessible in netstandard2.0, however [`ExceptionMethod` is initialized with `null` value](https://github.com/dotnet/coreclr/blob/526c32c221192cd5876239b2d72c68d1ba0c40c3/src/System.Private.CoreLib/src/System/Exception.cs#L447). So, there was introduced new compilation constant `PARTIALLY_SERIALIZATION` to handle such situation for `netstandard2.0`

### DIFF
--- a/src/Serilog.Formatting.Elasticsearch/Serilog.Formatting.Elasticsearch.csproj
+++ b/src/Serilog.Formatting.Elasticsearch/Serilog.Formatting.Elasticsearch.csproj
@@ -37,7 +37,7 @@
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
     <DefineConstants>$(DefineConstants);NO_SERIALIZATION</DefineConstants>
   </PropertyGroup>
-    
+
   <ItemGroup>
     <DotNetCliToolReference Include="dotnet-version" Version="1.1.0" />
   </ItemGroup>
--- a/src/Serilog.Formatting.Elasticsearch/Serilog.Formatting.Elasticsearch.csproj
+++ b/src/Serilog.Formatting.Elasticsearch/Serilog.Formatting.Elasticsearch.csproj
@@ -4,7 +4,7 @@
     <VersionPrefix>6.0.0</VersionPrefix>
     <VersionSuffix>alpha</VersionSuffix>
     <Authors>Michiel van Oudheusden, Martijn Laarman, Mogens Heller Grabe, Serilog Contributors</Authors>
-    <TargetFrameworks>net45;netstandard1.3</TargetFrameworks>
+    <TargetFrameworks>net45;netstandard1.3;netstandard2.0</TargetFrameworks>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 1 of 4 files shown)


mivano:
Thanks!

Issue #248, opened 2019-05-27
ElasicSearch 7.0 support


kanadaj:
ElasticSearch 7 isn't currently supported due to a change in how templates operate - the _default_ keyword is not accepted anymore.

The new template has to completely ditch the _default_ flag:

```cs
        private static object GetTemplateESv7(
            Dictionary<string, string> settings,
            string templateMatchString)
        {
            return new
            {
                index_patterns = new[] { templateMatchString },
                settings = settings,
                mappings = new
                {
                    dynamic_templates = new List<Object>
                    {
                        //when you use serilog as an adaptor for third party frameworks
                        //where you have no control over the log message they typically
                        //contain {0} ad infinitum, we force numeric property names to
                        //contain strings by default.
                        {
                            new
                            {
                                numerics_in_fields = new
                                {
                                    path_match = @"fields\.[\d+]$",
                                    match_pattern = "regex",
                                    mapping = new
                                    {
                                        type = "text",
                                        index = true,
                                        norms = false
                                    }
                                }
                            }
                        },
                        {
                            new
                            {
                                string_fields = new
                                {
                                    match = "*",
                                    match_mapping_type = "string",
                                    mapping = new
                                    {
                                        type = "text",
                                        index = true,
                                        norms = false,
                                        fields = new
                                        {
                                            raw = new
                                            {
                                                type = "keyword",
                                                index = true,
                                                ignore_above = 256
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    },
                    properties = new Dictionary<string, object>
                    {
                        {"message", new {type = "text", index = "true"}},
                        {
                            "exceptions", new
                            {
                                type = "nested",
                                properties = new Dictionary<string, object>
                                {
                                    {"Depth", new {type = "integer"}},
                                    {"RemoteStackIndex", new {type = "integer"}},
                                    {"HResult", new {type = "integer"}},
                                    {"StackTraceString", new {type = "text", index = "true"}},
                                    {"RemoteStackTraceString", new {type = "text", index = "true"}},
                                    {
                                        "ExceptionMessage", new
                                        {
                                            type = "object",
                                            properties = new Dictionary<string, object>
                                            {
                                                {"MemberType", new {type = "integer"}},
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            };
        }```

Also, upgrading from ES6 requires the deletion of the previous template:
```
DELETE  /_template/serilog-events-template```

[...]

PR #249, opened 2019-05-29
Elasticsearch.Net 7.0.0 support, V7 templates

kanadaj:
https://github.com/serilog/serilog-sinks-elasticsearch/issues/248 targeting Elasticsearch.Net 7.0.0-alpha2 on .NET Standard 2.0; still targeting 6.0.0 on .NET Standard 1.1 and .NET Framework 4.5.1 since 7.0.0 is not compatible.

A small note: why target .NET Framework separately?

Possibly. I cannot test against ES 6 right now so would need verification whether the 7.0.0 SDK of ElasticSearch breaks it or not.

**Other information**:
Needs testing against ES 6.X.


global.json
```diff
@@ -1,5 +1,5 @@
 {
     "sdk": {
-        "version": "2.1.4"
+        "version": "2.1.500"
```
ffMathy: Why 500 here?

kanadaj: @ffMathy oh that's just from my initial setup cause I don't have 2.1.4 locally and when I did that I was just experimenting, didn't consider I'd end up making a PR...

ffMathy: Ah I see.

dominiclapointe:
Is there a reason that it has not been merged yet? Really looking forward for the next nuget package that is compatible with esv7!!

kanadaj:
@dominiclapointe Yes, the fact that this pull request also contains a dependency on the Preview version of the SDK itself, and causes some other complications like leaving the dependency of the netcore1.3 package as 6.X and netcore2.0 as 7.X - they kind of have to figure out what to do with netcore1.3.

Also, I'm still not sure the preview SDK for 7.X is compatible at all with ES6

naymore:
Does anybody know what Elastic's idea is with its dotnet SDK?  
Is it backward compatible?  
Will 6.8 receive future updates (hotfix releases like 6.8.1, ..)? (In the past this almost never happened)  
Does the ES SDK for dotnet version match the supported ES version?  

ES v6 will be supported for quite some time, [see EOL](https://www.elastic.co/support/eol), however v7 introduced breaking changes and [v8 will introduce even more](https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking-changes-8.0.html).

btw, 7.0.0-beta1 has been released ;-)
Maybe this sink should also reflect the ES SDK used in its version scheme.
v7.1.0 sink only supports v6.8 ES SDK :|

It might be simpler to have ES-version-specific sinks (at least for ES v6.x and v7+) rather than having one sink for any ES version and also be in doubt what's even supported or breaking :)

mivano:
@Mpdreamz can you give some feedback here about the plans?

NeilCross:
Elasticsearch.net 7.0.0 has been released on nuget recently: https://www.nuget.org/packages/Elasticsearch.Net/

Mpdreamz:
I've started on a continuation of this PR here:

Issue #263, opened 2019-08-04
Exception while emitting periodic batch: System.Collections.Generic.Dictionary<string,object>' does not contain a definition for 'index'


krisztiankocsis:
**A few questions before you begin:**

**Does this issue relate to a new *feature* or an existing *bug*?**
- [X] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**

8.0.0 (ElasticSearch.Net 7.1.0)

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [X] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**
Continuously throwing exceptions in self-log.
Durable storage purge, still the same.

**Please describe the expected behavior?**
No exception should be thrown.

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**

Aug 04 18:01:45 cloud.mis-iwork.hu miw-cloud[22528]: 2019-08-04T16:01:45.9506278Z Exception while emitting periodic batch from Serilog.Sinks.Elasticsearch.Durable.ElasticsearchLogClient: Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: 'System.Collections.Generic.Dictionary<string,object>' does not contain a definition for 'index'
Aug 04 18:01:45 cloud.mis-iwork.hu miw-cloud[22528]:    at CallSite.Target(Closure , CallSite , Object )
Aug 04 18:01:45 cloud.mis-iwork.hu miw-cloud[22528]:    at System.Dynamic.UpdateDelegates.UpdateAndExecute1[T0,TRet](CallSite site, T0 arg0)
Aug 04 18:01:45 cloud.mis-iwork.hu miw-cloud[22528]:    at Serilog.Sinks.Elasticsearch.Durable.ElasticsearchLogClient.GetInvalidPayloadAsync(DynamicResponse baseResult, List`1 payload, List`1& cleanPayload)
[...]

PR #264, opened 2019-08-06
Changed property accessors to dictionary.

dpiessens:
It addresses #263 where exceptions keep getting thrown in the self log.

No, it should continue to work with older versions.

- [X ] The commit follows our [guidelines](https://github.com/serilog/serilog/blob/dev/CONTRIBUTING.md)
- [ ] Unit Tests for the changes have been added (for bug fixes / features) - No there is no testing for the durable section and I don't have domain knowledge of what client request/response looks like.

**Other information**:


mivano:
Sorry for the late response. I think this is now fixed in another PR already.

PR #266, opened 2019-08-20
Fixes two problems, needing to use indexers on a dynamic, and a logic…

matthewDDennis:
… error that prevented the durable log files from being processed.
This PR fixes two issues in ElasticsearchLogClient.cs.
- The first is similar to PR #264 to address the properties on a dymanic using index access instead of properties.  For some reason this property access doesn't work.
- The second is a logic error the stops the log files from being processed.


matthewDDennis:
The first problem is with how the code uses dynamic.
A dynamic variable is just something that can hold any type of object.  It is not automatically an ExpandoObject, which you can access with both indexer and property syntax.
In this case the type of the object in the dynamic variable is a Dictionary<string, object>, so only indexers can be used, not property syntax access.

mivano:
Thanks for the PR!

harlennaves:
When the will be the release of 8.1 version?

freeranger:
This seems not to be in 8.2 either - @mivano do you have an ETA for release of this fix?  Thanks

mivano:
Due to some versioning issue, this will be 8.4, which is just pushed to nuget.

PR #271, opened 2019-08-26
[Urgent] Fix KeyNotFoundException when `items[i].errors` is undefined

piaste:

Upon installing either the stable (8.0) or the preview (8.1 alpha 0001) version of this client, publishing to an ES 7 cluster, some of our logs would fail to appear in Kibana. Some of these were extremely plain `Log.Verbose("some literal string")` calls, which puzzled us a fair bit.

Enabling self-logging showed this exception:

```
exception.Message:The given key 'error' was not present in the dictionary.
exception.StackTraceString: at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
 at System.Dynamic.UpdateDelegates.UpdateAndExecute2[T0,T1,TRet](CallSite site, T0 arg0, T1 arg1)
 at Serilog.Sinks.Elasticsearch.ElasticsearchSink.EmitBatch(IEnumerable`1 events)
 at Serilog.Sinks.PeriodicBatching.PeriodicBatchingSink.EmitBatchAsync(IEnumerable`1 events)
 at Serilog.Sinks.PeriodicBatching.PeriodicBatchingSink.OnTick()
exception.ClassName:System.Collections.Generic.KeyNotFoundException exception.Source:System.Private.CoreLib
exception.RemoteStackTraceString: - 
exception.RemoteStackIndex:0 
exception.HResult:-2,146,232,969 
exception.HelpURL: - 
```

The stack trace led me to [this](https://github.com/serilog/serilog-sinks-elasticsearch/blob/9d9930e2e5e960dc8c1b230fc7eaaac5c84e3dc9/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs#L75) dictionary lookup.

I looked up a helper function to check dynamic objects for member existence, added the check, and with the modified .dll the missing log entries appear correctly.

No. The modified code gets called when `Errors` is set to `true` but some of the `items` don't have an `error` entry, and the current code appears to just abort in that case.

I didn't add new unit tests, in part because the change should be very safe, and in part because, as I hinted above, I'm not quite sure how to fully repro the original bug.


mivano:
Thanks for the quick fix. I see some other people struggling with this, so good to have this out.

piaste:
Addendum: I tagged this 'urgent' because we're going into prod soon, and it would be very appreciated if it could be inserted into an `-alpha0002` or something instead of having to ship a custom fork just for this purpose. Thanks :)

svenschelfaut:
When are you planning to release the stable 8.1 version?

mivano:
The 8.0.1 is released. I though I did this a week ago, but the build failed and I did not pay enough attention. Re-queued and is available now. Sorry for that.

svenschelfaut:
Thanks!

misterspeak:
> The 8.0.1 is released. I though I did this a week ago, but the build failed and I did not pay enough attention. Re-queued and is available now. Sorry for that.

Hi - it looks like this bug has reappeared in 8.1.0? Downgrading back to 8.0.1 removes the error, upgrade back to 8.1.0 and the error reappears.

Issue #278, opened 2019-09-27
minimumLogEventLevel setting does not work


SirSeven:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [ x ] Bug

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
8.0.0

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [ x ] netCore 2.0
- [ x ] 4.7

**Please describe the current behavior?**
When using `minimumLogEventLevel` via either xml or json based configuration (using Serilog.Settings.Configuration 3.1.0) the set value is neither checked nor used. All events are beeing sent.

**Please describe the expected behavior?**
The logged events should be checked if they are "higher" (more-important) than the given `minimumLogEventLevel`

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**
- create empty project
- install Serilog
- install Serilog.Sinks.Elasticsearch
- install Serilog.Settings.Configuration
- configure something like this:
`"Serilog": {
  "MinimumLevel": "Verbose",
  "WriteTo": [
    {
      "Name": "Elasticsearch",
      "Args": {
        "nodeUris": "http://someelastic",
        "indexFormat": "i-{0:yyyy.MM.dd}",
        "minimumLogEventLevel": "Warning",
      }
    }
  ]
}`
- write Log with log level verbose or information.
- you will find your logs in the elastic search

**Issue / Solution**
This seems to be just an documentation issue. The README.md indicates to use `minimumLogEventLevel` but the correct setting is `restrictedToMinimumLevel`. This is the issue to my pull-request.

mivano:
Merged, Thanks for the PR.

PR #279, opened 2019-09-27
Updated README.md

SirSeven:
#278 minimumLogEventLevel setting does not work 

No because if the wrong value was used before it wouldn't work anyway.

- [ x ] The commit follows our [guidelines](https://github.com/serilog/serilog/blob/dev/CONTRIBUTING.md)

**Other information**:
Documentation was describing a configuration value that was not used by the application. Which caused a unwanted behaviour. According to https://github.com/serilog/serilog-sinks-elasticsearch/blob/dev/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs#L96 and my tests the correct value is `restrictedToMinimumLevel`.


mivano:
Thanks for the PR!

Issue #281, opened 2019-10-10
Log Events not Written to Sink with Inner Exceptions


msfreed:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [x] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
8.0.0

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [ ] netCore 2.0
- [ ] netCore 1.0
- [x] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**
Log events are not being written to the configured sink/s if either the `ElasticsearchJsonFormatter` OR `ExceptionAsObjectJsonFormatter` are configured AND the Error level is specified AND an exception is specified with an inner exception AND `formatStackTrackAsArray` is set to `true`.

**Please describe the expected behavior?**
Log events with an inner exception would be written to the configured sink/s

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**

```
var logger = new LoggerConfiguration()
   .WriteTo.Console(new ElasticsearchJsonFormatter(formatStackTraceAsArray: true))
   .CreateLogger();
logger.Error(new Exception("outer", new Exception("inner")), "failure");
```

marcin-brzozowski:
This is also reproducible on .NET Core 2.x

mivano:
Merged to dev, so a package should be available soon.

PR #285, opened 2019-10-17
Fix log events not being written to Sink when exception has no stacktrace

marcin-brzozowski:
#281 - Log Events not Written to Sink with Inner Exceptions.

No

**Other information**:
A null reference exception was thrown when trying to format Exception with no stacktrace (either root or inner one). This caused log events not to be written to the Sink.

This only happens, when someone tries to log an exception that wasn't thrown, but created via the `new` keyword. The latter one has most properties set to `null` because it was not thrown. While it's rare that such instance of the Exception class is used, the formatter should still be able to handle it gently.

I think that when `formatStackTraceAsArray` is enabled, the `StackTrace` and `RemoteStackTrace` Exception properties, when `null`, should be formatted as empty arrays.


mivano:
Thanks for the PR!

Issue #294, opened 2019-11-10
An issue with BufferCleanPayload


daniel-oster:
Hey
I have an issue with the BufferCleanPayload in that its not called when we get an error from elastic :-)
and then there is a separate issue with the message provided is ToString of a dictionary, and hence just the type name...

A have a fix that I have tested locally for these issues and works fine as far as I can see and will cut a pr for it.

Thanks for a great component :-)
Cheers
 Daniel

**A few questions before you begin:**

> Is this an issue related to the [Serilog core project](https://github.com/serilog/serilog) or one of the [sinks](https://github.com/serilog/serilog/wiki/Provided-Sinks) or 
[community projects](https://github.com/serilog/serilog/wiki/Community-Projects).  
This issue list is intended for Serilog Elasticsearch Sink issues. If this issue relates to another sink or to the code project, 
please log on the related repository.  Please use [Gitter chat](https://gitter.im/serilog/serilog) and [Stack Overflow](http://stackoverflow.com/questions/tagged/serilog)  for discussions and questons.

**Does this issue relate to a new *feature* or an existing *bug*?**
- [ X] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
latest at least

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [x ] netCore 2.0
- [ x] netCore 1.0
- [ x] 4.7
- [ x] 4.6.x
- [ x] 4.5.x

**Please describe the current behavior?**
the BufferCleanPayload action is not called when it should, and if it was the error is the dictionary type name

**Please describe the expected behavior?**
that the cleanup action is called and and the error message contains the actual error message from elastic

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**
assign a action BufferCleanPayload and log something that is rejected (for example by having a conflicting type )

mivano:
Thanks, merged it to dev. Do use the alpha package to test and I will merge it soon to master for a final package.

PR #295, opened 2019-11-10
Fixed two issues regarding BufferCleanPayload

daniel-oster:
I fixed a logical test issue that made the code to not go through the
cleanup path when errors happened and a issue with the error message
where a ToString was called on a dictionary resulting in the type name
instead of the actual error message. I refrained from adding the
detailed errors provided by elastic, because they can be many and
ridiculously detailed

An issue with BufferCleanPayload #294

Not really, but since it may start to execute code paths not previously executed it could cause peoples bugs to surface 

- [ X] The commit follows our [guidelines](https://github.com/serilog/serilog/blob/dev/CONTRIBUTING.md)

**Other information**:


mivano:
Thanks for the PR

PR #298, opened 2019-11-22
Change default TypeName to '_doc'

### LINKED ISSUE #200: Change default TypeName to _doc
Hello,

Change the default TypeName to _doc because the [mapping types are removed](https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html) in the future version of Elasticsearch (7.0).

Sorry it's not a bug nor a feature but I don't know where to post..

Excolo:
Fixes #200

It introduces a potential "breaking change", since it influences code relying on the former default `TypeName`.

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkOptions.cs
+++ b/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkOptions.cs
@@ -278,7 +278,7 @@ namespace Serilog.Sinks.Elasticsearch
         /// The default Elasticsearch type name used for Elasticsearch versions prior to 7.
         /// <para>As of <c>Elasticsearch 7</c> and up <c>_type</c> has been removed.</para>
         /// </summary>
-        public static string DefaultTypeName { get; } = "logevent";
+        public static string DefaultTypeName { get; } = "_doc";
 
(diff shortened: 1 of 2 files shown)


mivano:
Thanks for the PR, this is indeed a breaking change. But when needed at some point, we should go for it. I m not completely into this anymore, @Mpdreamz can I ask you for some advice? Would this be needed already, be conditional (as we are supporting now some concept of ES versions) or break and document?

Mpdreamz:
The current form its conditional:

If  `_options.DetectElasticsearchVersion`  is set and we see that we are targeting `ES 7.x` we default to `_doc`. 

We also detect if the intended template to be registered is the v6 to include type name or not in the put template call.

`DetectElasticsearchVersion` defaults to false though as to not need to introduce a new major version. 


I think defaulting to `_doc` and possibly enabling `DetectElasticsearchVersion` by default would be good at this point in time to move forward.

mivano:
Thanks for the input. @Excolo can you reflect this in the readme and changelog markdown files as well?

Excolo:
@mivano Certainly. I'll see to it tomorrow.

mivano:
@Excolo had any time to change it in the docs?

mivano:
This is merged into dev now.

mivano:
Thanks for the PR. I updated the changelog.

PR #300, opened 2019-12-02
Fix/dev/cleanup

Mpdreamz:
This mainly makes the TargetFramework selection dependent on the operating system.

It also moves the .NET sdk over to 3.0.


mivano:
Since this one is too old, I m closing it.

PR #320, opened 2020-03-05
Allow overriding elasticsearch logclient

vilppu:
Allow ElasticsearchLogClient factory to be passed as option for DurableElasticsearchSink

The `ElasticsearchLogClient` uses internally `DynamicResponse` object that is not usable in when application is compiled to native code. This is usually the case with UWP apps as those are compiled to native code by default on release mode.

So this change allows `ElasticsearchLogClient` to injected to `DurableElasticsearchSink` so that the client can be overridden with custom implementation when the default `ElasticsearchLogClient` cannot be used.

For more information about the issue with `DynamicResponse` and native code compilation see elastic/elasticsearch-net#4389


russcam:
Another issue is that the internal serializer uses a modified version of Utf8Json, which uses Reflection.Emit to generate IL for serialization routines. There are plans to move to System.Text.Json in the future, but as far as I am aware, it also uses Reflection.Emit to generate IL at runtime.

To allow the sink to run universally, you _may_ want to use a JSON serializer that will run on UWP, like Json.NET, which will fall back to expression API and cached reflection routines when Reflection.Emit is not supported. A serializer based on Json.NET can be hooked up with the low level client by passing an implementation of `IElasticsearchSerializer` to `ConnectionConfiguration`. If using Json.NET, could then use an implementation of `ElasticsearchResponse<JObject>` as the response.

vilppu:
> Another issue is that the internal serializer uses a modified version of Utf8Json, which uses Reflection.Emit to generate IL for serialization routines. There are plans to move to System.Text.Json in the future, but as far as I am aware, it also uses Reflection.Emit to generate IL at runtime.
> 
> To allow the sink to run universally, you _may_ want to use a JSON serializer that will run on UWP, like Json.NET, which will fall back to expression API and cached reflection routines when Reflection.Emit is not supported. A serializer based on Json.NET can be hooked up with the low level client by passing an implementation of `IElasticsearchSerializer` to `ConnectionConfiguration`. If using Json.NET, could then use an implementation of `ElasticsearchResponse<JObject>` as the response.

Good point! That is exactly what we had to do in the UWP project.

Our application has been working nicely with  custom `IElasticsearchSerializer`  when the changes that are in this PR were applied to elastic sink.

mivano:
Thanks for the PR and sorry for not coming back to you faster! Is this still an issue you would like to solve using the code proposed in this PR? My only concern is that we are adding more and more options and extensibility points to the sink. Not directly bad, but harder to find and use. Would it be possible to add some documentation of this in the readme.md file for example?

vilppu:
> Thanks for the PR and sorry for not coming back to you faster! Is this still an issue you would like to solve using the code proposed in this PR? My only concern is that we are adding more and more options and extensibility points to the sink. Not directly bad, but harder to find and use. Would it be possible to add some documentation of this in the readme.md file for example?

Sorry I had missed this comment. I am not currently working on the project anymore where this was needed so I cannot say if this still needed. Maybe I can just close this PR to avoid unnecessary complexity in code base :)

I will mentions this to people who are still working with code that had this issue so they can reopen this if needed.

Issue #340, opened 2020-06-30
AutoRegisterTemplate fails on full .NET Framework


uvw:
The issue is related to the Serilog Elasticsearch Sink.

**Does this issue relate to a new *feature* or an existing *bug*?**
- [x] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**

https://www.nuget.org/packages/Serilog.Sinks.ElasticSearch/8.1.0 and likely any version older.

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [ ] netCore 2.0
- [ ] netCore 1.0
- [x] 4.7
- [x] 4.6.x
- [x] 4.5.x
- [x] Mono 6.10.x

**Please describe the current behavior?**

AutoRegisterTemplate fails on full .NET Framework. SelfLog indicates an exception involving 404 or 200 response codes even though the sink is correctly expects them (`AllowedStatusCodes = new[] { 200, 404 }`):

```
Failed to create the template. Elasticsearch.Net.ElasticsearchClientException: Request failed to execute. Call: Status code 200 from: HEAD /_template/logs-live-test-template
  at Elasticsearch.Net.Transport`1[TConnectionSettings].HandleElasticsearchClientException (Elasticsearch.Net.RequestData data, System.Exception clientException, Elasticsearch.Net.IElasticsearchResponse response) [0x00064] in <28cd31c834d342448debe9682d4bf644>:0
  at Elasticsearch.Net.Transport`1[TConnectionSettings].FinalizeResponse[TResponse] (Elasticsearch.Net.RequestData requestData, Elasticsearch.Net.IRequestPipeline pipeline, System.Collections.Generic.List`1[T] seenExceptions, TResponse response) [0x0004b] in <28cd31c834d342448debe9682d4bf644>:0
  at Elasticsearch.Net.Transport`1[TConnectionSettings].Request[TResponse] (Elasticsearch.Net.HttpMethod method, System.String path, Elasticsearch.Net.PostData data, Elasticsearch.Net.IRequestParameters requestParameters) [0x00197] in <28cd31c834d342448debe9682d4bf644>:0
[...]

PR #343, opened 2020-07-08
Upgrade Elasticsearch.Net to 7.8.1 

uvw:

Issue #340.

None in Serilog.Sinks.Elasticsearch.

There were a few breaking changes in Elasticsearch.Net between v7.0.0–v7.8.1 but the sink is not affected by them as far as I could tell and test.

**Other information**:

I had to introduce two extra changes for the upgrade:

1. Upgrade `Microsoft.CSharp` v4.5.0 → v4.6.0 because it was upgraded in Elasticsearch.Net as well.
1. Replace `Elastic.Managed` and `Elastic.Xunit` packages in the integration tests with the newer `Elastic.Elasticsearch.Managed` and `Elastic.Elasticsearch.Xunit`. The old packages were deleted from the Elastic Abstractions CI feed and not available anymore.

I ran the existing unit and integration tests (ES v6.0 and v7.8) on net461 (mono 6.10.0) and netcoreapp 2.1. I was not able to test on Windows since I don't have one atm. Hopefully, they will pass in CI on Windows 🙏


mivano:
Thanks for the PR!

PR #351, opened 2020-07-30
Issue nr: 14594 - removing old implementation where fields where dot-…

rajmondburgaj:
…escaped due to old ELK not supporting it. But now this is supported, therefore there is no need to escape it anylonger.

Disable dot escaping for field names, as it is supported since long time ago in ELK.

If the field name `x.y` was used, till now in kibana it would show as `x/y`, after this commit, it will show as `x.y`. But I think the breaking change is very small.

There was no need for new unit tests, since only old code was removed.

**Other information**:

Also changed a method from using recursion to a simple while loop.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ## Changelog
 
+* Disable dot-escaping for field names, because ELK already supports dots in field names.
+
 8.2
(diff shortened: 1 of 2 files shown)


mivano:
Thanks for the PR and sorry for not responding sooner!

Issue #355, opened 2020-09-11
Support for data streams Elastic 7.9


orjan:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [x] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
`<PackageReference Include="serilog.sinks.elasticsearch" Version="8.2.0" />`

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [x] netCore 3.1.402

**Please describe the current behavior?**
[Data streams](https://www.elastic.co/guide/en/elasticsearch/reference/master/data-streams.html) is a reliable way to minimize the number of shards in a cluster. Previously you had to setup an ILM, template and a write alias, data streams is more or less the same but endorsed by Elastic an transparent in Kibana and easier to configure.

There a constraint within data stream and it will only support the `op_type=create`
https://www.elastic.co/guide/en/elasticsearch/reference/master/use-a-data-stream.html#data-streams-bulk-indexing-requests

When trying to index documents from serilog the request looks like this
```
POST _bulk
{"index":{"_index":"my-data-stream","_type":"_doc"}}
{"@timestamp":"2020-09-11T16:14:17.6858199+02:00","level":"Error","messageTemplate":"Hello data stream from Serilog","message":"Hello data stream from Serilog"}
```

```
#! Deprecation: [types removal] Specifying types in bulk requests is deprecated.
{
  "took" : 0,
  "errors" : true,
  "items" : [
    {
      "index" : {
        "_index" : "my-data-stream",
        "_type" : "_doc",
        "_id" : null,
        "status" : 400,
        "error" : {
          "type" : "illegal_argument_exception",
          "reason" : "only write ops with an op_type of create are allowed in data streams"
        }
      }
    }
  ]
}
```

**Please describe the expected behavior?**
In order to support data streams we'll need to be able to alter the `op_type` from `index` to `create` then it should work:
https://github.com/serilog/serilog-sinks-elasticsearch/blob/master/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs#L156-L160

There's also a deprecation warning for the `_type` removal but that's covered by a separate issue: #345

From an API point of view I can imagine specifying the op_type similar to this?
```
[...]

Issue #345, opened 2020-07-11
HTTP 400 when sending log to ElasticSearch 8.0.0


Belorus:
.NET Core 3.1
Serilog 2.9.0
Serilog.Sinks.ElasticSearch 8.1.0

ElasticSearch 8.0

![image](https://user-images.githubusercontent.com/1737207/87233298-c2f90380-c3ce-11ea-9ac5-13b5dab3b2b0.png)

mivano:
Thanks for reporting. I have not used this version yet. Do you have a proposal on how to tackle this?

orjan:
One way could possible be to specify the Elastic version in the configuration. That is easy to understand but make it mandatory would cause a breaking change.

Making it optional and detect the version if not specified by calling `GET /` could also be an option to set the parameter if not specified?

Another option would be to have a default version and if possible configure the version in runtime?
https://github.com/uken/fluent-plugin-elasticsearch#verify_es_version_at_startup

mivano:
We do have an detection of the version, so we can use that?

orjan:
@Belorus would you mind try to set `TypeName = null` in the latest version `8.4.0`

https://github.com/serilog/serilog-sinks-elasticsearch/blob/dev/CHANGES.md

hastarin:
Silly me just updated our dev ELK stack to 8.0.0 without checking things would work.

I see this is also tracked via #375 

At the moment it looks like updating all our apps configuration via code will be needed:
`                                          TypeName = null,
                                          BatchAction = ElasticOpType.Create`

digger73:
I also ran into this problem and I'm reliant on the config files. However, it isn't possible to set `TypeName` to null. Is it possible to combine the use of configuraiton files and code config ?

orjan:
@digger73 I'm not running Elastic any longer but if I'm not wrong, it must be set through code since it's not possible to deliberately set it to `null` trough configuration?

digger73:
@orjan I came to the same conclusion, that was looking for the solution of using **both** the app.config and configuration of the sink in the code, however I haven't gotten this to work. When using app.config the elasticsearch sink is registered as a restricted sink and the options set through code is registered as an elasticsearch sink.

PR #356, opened 2020-09-13
Added support for controlling the op_type when indexing

orjan:
In order to write to data streams we'll need to set the
`op_type` to `create` the default is `index`

In order to don't serialize null values for the batch action and to not depend
on custom serialization setting we're using `LowLevelRequestResponseSerializer.Instance`
for serialization of the action.

We won't throw exception when setting the TypeName to null since:
- This is deprecated in v7
- Won't work in v8

- #355 It's possible to set the op_type to create by setting `BatchAction` to `ElasticOpType.Create`
- #345 It's possible to set `TypeName` to `null` to remove it from the payload being sent to Elastic.

I don't think we have introduced an breaking change but we'll need to review the following changes to make sure they're not breaking.

- The custom serializer for the action has been replaced by a specific one, it's not impossible that someone has been intercepting this call for adding custom elements when logging.
https://github.com/serilog/serilog-sinks-elasticsearch/pull/356/files#diff-15ec7bbc1a4d48125fae012ffb1896acR191

- Earlier and exception was thrown when trying to set the `TypeName` to `null` it's unlikely that someone has relied on this exception. https://github.com/serilog/serilog-sinks-elasticsearch/pull/356/files#diff-cb61ceb2dfe8091b1abe76e3c275d6d0L77

**Other information**:

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
+++ b/src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
@@ -143,6 +143,7 @@ namespace Serilog
         /// <param name="failureSink">Sink to use when Elasticsearch is unable to accept the events. This is optionally and depends on the EmitEventFailure setting.</param>   
         /// <param name="singleEventSizePostingLimit"><see cref="ElasticsearchSinkOptions.SingleEventSizePostingLimit"/>The maximum length of an event allowed to be posted to Elasticsearch.default null</param>
         /// <param name="templateCustomSettings">Add custom elasticsearch settings to the template</param>
+        /// <param name="batchAction">Configures the OpType being used when inserting document in batch. Must be set to create for data streams.</param>
         /// <returns>LoggerConfiguration object</returns>
         /// <exception cref="ArgumentNullException"><paramref name="nodeUris"/> is <see langword="null" />.</exception>
         public static LoggerConfiguration Elasticsearch(
@@ -180,7 +181,8 @@ namespace Serilog
             ILogEventSink failureSink = null,
             long? singleEventSizePostingLimit = null,
             int? bufferFileCountLimit = null,
-            Dictionary<string,string> templateCustomSettings = null)
+            Dictionary<string,string> templateCustomSettings = null,
+            ElasticOpType batchAction = ElasticOpType.Index)
         {
             if (string.IsNullOrEmpty(nodeUris))
                 throw new ArgumentNullException(nameof(nodeUris), "No Elasticsearch node(s) specified.");
@@ -208,6 +210,7 @@ namespace Serilog
             }
 
             options.BatchPostingLimit = batchPostingLimit;
+            options.BatchAction = batchAction;
             options.SingleEventSizePostingLimit = singleEventSizePostingLimit;
             options.Period = TimeSpan.FromSeconds(period);
             options.InlineFields = inlineFields;
--- a/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs
+++ b/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs
@@ -16,9 +16,9 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Elasticsearch.Net;
-using Elasticsearch.Net.Specification.SecurityApi;
 using Serilog.Debugging;
 using Serilog.Events;
 using Serilog.Sinks.PeriodicBatching;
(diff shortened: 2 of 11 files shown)


mivano:
Thanks for this PR. Looks good. I agree with your points, the exception type does not look like an issue, the other one might be an issue. But that we can highlight in the changes.md and readme.md.

orjan:
@mivano currently there's an issue with the error handling since when we're setting the `OpType` to `create` it won't work, since it's looking for `index`. Also there an open PR #284 trying to address another issue with error.

The question is would we like to stay with the DynamicResponse or are we open to de-serialize the response to DTO:s instead?

src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
```diff
@@ -143,6 +143,7 @@ public static class LoggerConfigurationElasticsearchExtensions
         /// <param name="failureSink">Sink to use when Elasticsearch is unable to accept the events. This is optionally and depends on the EmitEventFailure setting.</param>   
         /// <param name="singleEventSizePostingLimit"><see cref="ElasticsearchSinkOptions.SingleEventSizePostingLimit"/>The maximum length of an event allowed to be posted to Elasticsearch.default null</param>
         /// <param name="templateCustomSettings">Add custom elasticsearch settings to the template</param>
+        /// <param name="batchAction">Configures the OpType being used when inserting document in batch. Must be set to create for data streams.</param>
```
orjan, line 146: - [ ] Is `batchAction` a good name or should we call it `bulkAction`?

src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs
```diff
@@ -261,5 +255,45 @@ private void HandleResponse(IEnumerable<LogEvent> events, DynamicResponse result
                 HandleException(result.OriginalException, events);
             }
         }
+        
+        sealed class ElasticCreateAction
```
orjan, line 259: - [ ] Should we move the DTO:s to a separate file?

src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
mivano: Good question; Batch is used inside the sink, while Bulk is the actual command used to communicate to ES. As this is a setting specific to the Elasticsearch operation, it might be better to call it bulkAction instead.

src/Serilog.Sinks.Elasticsearch/LoggerConfigurationElasticSearchExtensions.cs
```diff
@@ -180,7 +181,8 @@ public static class LoggerConfigurationElasticsearchExtensions
             ILogEventSink failureSink = null,
             long? singleEventSizePostingLimit = null,
             int? bufferFileCountLimit = null,
-            Dictionary<string,string> templateCustomSettings = null)
+            Dictionary<string,string> templateCustomSettings = null,
+            ElasticOpType batchAction = ElasticOpType.Index)
```
mivano, line 185: Since this is a binary breaking change, we need to clearly indicate this change in the changes.md

src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs
mivano: as they are very specific to this implementation and internal, I would be fine keeping them here.

src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkState.cs
```diff
@@ -74,10 +74,6 @@ private ElasticsearchSinkState(ElasticsearchSinkOptions options)
             {
                 options.TypeName = "_doc";
```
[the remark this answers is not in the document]
mivano: My use of ES is at the moment pretty low, so I cant hardly see what the impact is and if we even need a template for v8. What is your opinion?

mivano:
Another thing that crossed my mind; the durable part also sends data to ES in a similar manner, but using different code constructs. Most likely the same issue will apply here. For some while now I feel less comfortable with this durable part. It is a large part of the code base with a lot of duplication. And that last part makes introducing these changes more difficult. Besides, the durable functionality is not as stable as it should be and there might be better solutions if you want similar behaviour. 

What is your take on this? Worthwhile to apply the same thing to the durable part, exclude it explicitly or something else?

src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkState.cs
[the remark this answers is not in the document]
orjan: Templates in `v7.9` and `v8` has been rewritten to work with composition instead for inheritance so they are different.
I'll suppose the current v7 templates will work in v8 as well but I haven't tested.
Implementation of the new templates is too big to fit in this merge request.

The current code in this merge request is setting `_doc` as default if nothing else is specified:
- null is allowed if specified
- we won't override `foobar` with `_doc` if we're setting the template.

If it's not enough with a change log, we'll need to introduce a new boolean field like this:
https://github.com/uken/fluent-plugin-elasticsearch#suppress_type_name

I'm open for both solutions!

orjan:
> Another thing that crossed my mind; the durable part also sends data to ES in a similar manner, but using different code constructs. Most likely the same issue will apply here. For some while now I feel less comfortable with this durable part. It is a large part of the code base with a lot of duplication. And that last part makes introducing these changes more difficult. Besides, the durable functionality is not as stable as it should be and there might be better solutions if you want similar behaviour.
> 
> What is your take on this? Worthwhile to apply the same thing to the durable part, exclude it explicitly or something else?

I did a quick fix for the durable part and there are 2 places that needs to be fixed, so I'll fix that as well.
I'm not sure what we should do about the duplicated implementations, in the best of world we should isolate the durable part to a separate lib or at least namespace if possible? Are there any other durable sinks with the same needs?

mivano:
A totally separate discussion of course from this PR, but I would be inclined to remove it out of this sink and offer functionality in a different way. Preferable using a higher abstraction, so you can specify it on the serilog sinks level instead. This original implementation came from the Seq one (https://github.com/serilog/serilog-sinks-seq/tree/dev/src/Serilog.Sinks.Seq/Sinks/Seq/Durable) so there are more sinks with a similar feature. Problem is that I do not know how many users are actually depending on this durable feature. So it would be a real breaking change with or without another solution.

src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkState.cs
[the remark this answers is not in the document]
mivano: Specifying a null and good changelog would be okay I think.

mivano (approved):

mivano:
Merging it into dev so we have a prerelease package.

Issue #357, opened 2020-09-16
Serilog.Sinks.Elasticsearch don't call any error handlers when logs are rejected


Evolva:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [X] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
8.2.0 

**Please describe the current behavior?**
When an LogEvent is rejected from ElasticSearch none of the configure handlers are called

**Please describe the expected behavior?**
Configured handlers should be called

**Analysis**
After some digging it seems that: 
https://github.com/serilog/serilog-sinks-elasticsearch/blob/1cf2748b7a85fdd95b2c863ed7d5c54a2252a78c/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs#L224-L233

don't detect my error because `settings` is already a `Dictionary<string, object>`
![issue_ES_Sink](https://user-images.githubusercontent.com/2643865/93337557-97a8ef80-f829-11ea-8c65-6f5086e6dbfd.png)

I rebuild a new version of the sink with 
```cs
private static bool HasProperty(dynamic settings, string name)
{
    if (settings is IDictionary<string, object>)
        return ((IDictionary<string, object>)settings).ContainsKey(name);
    
    if (settings is System.Dynamic.DynamicObject)
        return ((System.Dynamic.DynamicObject)settings).GetDynamicMemberNames().Contains(name);

    return settings.GetType().GetProperty(name) != null;
}
```
and it worked.

mivano:
Thanks for raising this and providing a PR! It is merged to master, trying to collect some more needed changes and bring a package out.

PR #358, opened 2020-09-16
Fix HasProperty when settings is a Dictionary

Evolva:
#357

No, ExpandoObject is already an IDictionary<string, object>


orjan:
Looks great!
Also, you're not alone, there are a couple of pull requests that's trying to addressing this issue:
- #284
- #355

-

mivano:
Thanks for the PR!

Issue #316, opened 2020-02-12
exception is occuring when elastic server is missing


Su-s:
Below is the code for serilog elastic search.It's working perfectly when elastic service is running but fails when the server is not available

```
  .WriteTo.Elasticsearch(new ElasticsearchSinkOptions(new Uri("http://localhost:9200/")) // humio elastic bulk endpoint
                            {
                                AutoRegisterTemplate = true,
                                OverwriteTemplate = true,
                                DetectElasticsearchVersion = true,
                                AutoRegisterTemplateVersion = AutoRegisterTemplateVersion.ESv7,
                                NumberOfReplicas = 1,
                                NumberOfShards = 2,
                                RegisterTemplateFailure = RegisterTemplateRecovery.FailSink,
                                FailureCallback = e => Console.WriteLine("Unable to submit event " + e.MessageTemplate),
                                EmitEventFailure = EmitEventFailureHandling.WriteToSelfLog |
                                                   EmitEventFailureHandling.WriteToFailureSink |
                                                   EmitEventFailureHandling.RaiseCallback,
                                //FailureSink = new FileSink("./fail-{Date}.txt", new JsonFormatter(), null, null)
                                //ModifyConnectionSettings = x => x.BasicAuthentication(username: "", password: "88WS28JEuB0G2WC3cYufdgTINAxOGQvliBrqH5Vqutjb")  // password is ingest token from humio
                            })
```

exception is occuring when elastic server is missing

Elasticsearch.Net.ElasticsearchClientException: 'No connection could be made because the target machine actively refused it.. Call: Status code unknown from: GET /_cat/nodes?h=v'
 
    HttpRequestException: No connection could be made because the target machine actively refused it.
    SocketException: No connection could be made because the target machine actively refused it.
 
    This exception was originally thrown at this call stack:
    System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
    System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
    System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
    System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
    System.Net.Http.ConnectHelper.ConnectAsync(string, int, System.Threading.CancellationToken)

mivano:
[...]

PR #359, opened 2020-09-16
Make sure the sink (and app) do not crash when the ES is unreachable …

mivano:
Make sure the sink (and app) do not crash when the ES is unreachable during the discovery of the ES version.

Fixes the problem raised in this issue: https://github.com/serilog/serilog-sinks-elasticsearch/issues/316

If you are now depending on an app crash when the ES is not available during detection of the version, then this will no longer be the case. The error is logged to the selflog, but a sink should not bring down an app.

- [X] The commit follows our [guidelines](https://github.com/serilog/serilog/blob/dev/CONTRIBUTING.md)
- [X] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:


orjan:
It's good that the application not crashes, still the behavior isn't deterministic.
That make we wonder if we'll need dynamic looks up in future versions?
- Running an application against a specific version of Elastic is often a deliberate choice.
- Make the sink less complex and deterministic
- Changes are typically gradually introduced e.g. in the current versions of `v7` it's possible to drop the `_type`

Maybe we should move this discussion to a separate issue since it's not directly related to the pull request, that looking good.

mivano:
Agreed, it would be strange to have a suddenly changed version of Elasticsearch. The original intent was to make it as easy as possible to get started with the default settings. So no manual template and mapping setup. However, we are now supporting too many options and versions, making it more complex. Combined with a lot of duplicated code for the buffer option, we might need to reconsider some options here.

Issue #364, opened 2020-09-25
Upgrade issues from 8.2.0 to 8.4.0 (illegal_argument_exception)


estateretailsolutions:
**A few questions before you begin:**

> Is this an issue related to the [Serilog core project](https://github.com/serilog/serilog) or one of the [sinks](https://github.com/serilog/serilog/wiki/Provided-Sinks) or 
[community projects](https://github.com/serilog/serilog/wiki/Community-Projects).  
This issue list is intended for Serilog Elasticsearch Sink issues. If this issue relates to another sink or to the code project, 
please log on the related repository.  Please use [Gitter chat](https://gitter.im/serilog/serilog) and [Stack Overflow](http://stackoverflow.com/questions/tagged/serilog)  for discussions and questions.

This is related to either Serilog.Formatting.Elasticsearch OR Serilog.Sinks.Elasticsearch

**Does this issue relate to a new *feature* or an existing *bug*?**
- [X] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**

8.4.0

Serilog.Sinks.Elasticsearch 8.4.0
Serilog.Formatting.Elasticsearch 8.4.0

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [X] netCore 3.1
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**

We upgraded our nuget packages from 8.2.0 to 8.4.0 and the seriloging to ES stopped working.
We now receive this message in our console:

2020-09-25T14:54:58.8381985Z Failed to store event with template 'XXXX' into Elasticsearch. Elasticsearch reports for index logstash-2020.09.25 the following: {"type":"illegal_argument_exception","reason":"Rejecting mapping update to [logstash-2020.09.25] as the final mapping would have more than 1 type: [_doc, logevent]"}

We have multiple services that utilize this logging setup and it works just fine with 8.2.0 but does not with 8.4.0.
No code changes on our side.  We are hitting a local Elasticsearch instance running on windows.
elasticsearch 7.9.2

**Please describe the expected behavior?**

I would expect it to continue logging to elastic search as it was with 8.2.0.

[...]

PR #365, opened 2020-09-26
Override type name if specified

orjan:
Fixes a regression introduced in #356
We should override the TypeName if it's specified.

Note that there are different defaults depending
on how the options are created.
- Created from configuration it defaults to `logevent`
- Created from code it defaults to `_doc`

#364

No


mivano:
Thanks for looking into this and providing a fix!

Issue #375, opened 2020-12-11
Error when sink is connected to ElasticSearch 8.0.0-SNAPSHOT


thomashoefkens-vw:
**A few questions before you begin:**

> Is this an issue related to the [Serilog core project](https://github.com/serilog/serilog) or one of the [sinks]
Sink Elasticsearch

**Does this issue relate to a new *feature* or an existing *bug*?**
- [X] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
8.4.1

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**
- [X] netCore 5.0

**Please describe the current behavior?**
When connecting to Elastic 7.9.1 logging works fine. Connecting to 8.0 throws this error:
    
```
2020-12-11T12:56:03.7531402Z Caught exception while preforming bulk operation to Elasticsearch: Elasticsearch.Net.ElasticsearchClientException: Request failed to execute. Call: Status code 400 from: POST /_bulk. ServerError: Type: illegal_argument_exception Reason: "Action/metadata line [1] contains an unknown parameter [_type]"
11.12.2020, 13:56:03
orderservice-api-745b5975df-2jcbm
b986a1e1f9483f1c4ce20673c607f71eb3608b85ee9b1882f6a0eab169f3a594
at Elasticsearch.Net.Transport`1.HandleElasticsearchClientException(RequestData data, Exception clientException, IElasticsearchResponse response)
020, 13:56:03
orderservice-api-745b5975df-2jcbm
b986a1e1f9483f1c4ce20673c607f71eb3608b85ee9b1882f6a0eab169f3a594
at Elasticsearch.Net.Transport`1.FinalizeResponse[TResponse](RequestData requestData, IRequestPipeline pipeline, List`1 seenExceptions, TResponse response)
11.12.2020, 13:56:03
orderservice-api-745b5975df-2jcbm
b986a1e1f9483f1c4ce20673c607f71eb3608b85ee9b1882f6a0eab169f3a594
at Elasticsearch.Net.Transport`1.RequestAsync[TResponse](HttpMethod method, String path, CancellationToken cancellationToken, PostData data, IRequestParameters requestParameters)
11.12.2020, 13:56:03
[...]

PR #420, opened 2022-04-20
Proper handling of TypeName = null from appsettings.json

mariwing:
Ability to set TypeName to null via appsettings.json. (Root cause of issue #375 )

No

Unable to pass most of the unit tests as they seem to need a real cluster to run towards?

**Other information**:
TypeName was already exposed as configurable through appsettings.json but if set to null it is (by Microsofts configuration reader I belive) set to an empty string. This in turns make the already exisiting handling of TypeName in the sink (TypeName is deprecated in Elastic 7.x and unsopported (posts get 400 reply) in Elastic 8.x) not work as intentended.


cbb3268665 (approved):

johnarsc (approved):

mivano:
Thanks for the PR. I merged it into dev so it will get a beta package.

Issue #381, opened 2021-02-02
FailureCallback does not expose the Exception that caused the failure


staff0rd:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [x] New Feature

**Please describe the current behavior?**
`FailureCallback` does expose the `Exception` causing the failure

**Please describe the expected behavior?**
`FailureCallback` could expose the `Exception` that caused the failure

I came across this myself, and then saw a [similar question on Stackoverflow](https://stackoverflow.com/questions/65081219/elk-serilog-sink-failurecallback/65979206#65979206). Is there any appetite for merging something that would expose the exception, [for example, something like this](https://github.com/serilog/serilog-sinks-elasticsearch/commit/9b90b50be256a25c5f221c492669304a313b1950)?

mivano:
The purpose is to somehow store the lost events to some other location, not directly handle the underlying exception. That exception will be the same for all the event that suffered this problem while being stored to Elasticsearch. Like a connection exception. 
What is your use case that you want to handle this exception as well?

jdiosdado:
> The purpose is to somehow store the lost events to some other location, not directly handle the underlying exception. That exception will be the same for all the event that suffered this problem while being stored to Elasticsearch. Like a connection exception.
> What is your use case that you want to handle this exception as well?

There can be multiple causes on why the logs are not being stored, how would someone fix the issue if there is no error data?
You are right, the exception will be the same but in order to research how to fix one needs to know what the actual issue is.

bferdinandus:
+1
I would also like to see the exeption in the failure callback. That would simplify the debugging process as to discover why an event could not be send to elastic. Now I had to make use of a web debugging proxy (Charles)

iamitdubey:
+1
Need this to be able to know the reason for failed events

Towmeykaw:
My reason for wanting this is to be able to filter out the failed events which needs to be handled. Like mapper_parsing_exception which I would like to escalate but then ignore stuff like connection errors

Towmeykaw:
@mivano Is there a chance to add this feature?

mivano:
Looks like something useful based on the feedback. Care for a PR?

anujacharya997:
> 

is this aviable now ? i am using Serilog.Sinks.Elasticsearch, Version=8.4.1.0 in which version will this be aviable

Rombersoft:
[...]

PR #449, opened 2022-06-13
Added original exception to FailureCallback action

Towmeykaw:
#381 

Yes

**Other information**:
At first I thought it would be nice to populate the Exception in the Logevent but unfortunately it had no setter. So had to be a breaking change instead

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs
+++ b/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs
@@ -139,7 +139,7 @@ namespace Serilog.Sinks.Elasticsearch
                 {
                     foreach (var e in events)
                     {
-                        _state.Options.FailureCallback(e);
+                        _state.Options.FailureCallback(e, ex);
                     }
                 }
                 catch (Exception exCallback)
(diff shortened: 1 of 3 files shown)


AlexGoris-KasparSolutions:
Can we get this merged, I need exactly this  to investigate why I get elastic failures every now and then :)

src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticSearchSink.cs
```diff
@@ -242,7 +242,7 @@ private void HandleResponse(IEnumerable<LogEvent> events, DynamicResponse result
                             // Send to a failure callback
                             try
                             {
-                                _state.Options.FailureCallback(e);
+                                _state.Options.FailureCallback(e, null);
```
AlexGoris-KasparSolutions, line 245: We should find a way to raise a meaningful exception here as well. I did a test with 403 response from elastic and ended up here but the `null` doesn't tell anything to the higher-up code.
This gives me usable information to act upon (copied selflog from line 216):
```csharp
 _state.Options.FailureCallback(e, new Exception(string.Format("Failed to store event with template '{0}' into Elasticsearch. Elasticsearch reports for index {1} the following: {2}",
                                e.MessageTemplate,
                                action["_index"],
                                _state.Serialize(action["error"]))));
```

netclectic:
Is this ever going to get merged?
It seems like a massive oversight that this wasn't included from day 1.

PR #451, opened 2022-06-27
default typeName is null since _type is removed in ESv7 and unsupported in ESv8+

patrikbeno:
…in ESv8+

#375 

Technically, yes, but since _type is unsupported in ESv8+, then NO. You can't break what is already broken.

**Other information**:

current default is a blocker when trying to configure ES sink from json/yaml, since typename=null or typename="" is just ignored and default typename="_doc" remains in place


nenadvicentic:
I am also having issues with `Serilog.Sinks.Elasticsearch` to Elasticsearch 8.x. I would just love that, whatever implementation goes forward to fix the issue with `8.x` does not break `7.x`. We have enough problems using this package already.

PR #459, opened 2022-08-16
Support read formatStackTraceAsArray  from configuration

AiSY-Yang:
Support read formatStackTraceAsArray  from configuration

When I Add `formatStackTraceAsArray` in my configuration file ,this paramater does not take effect.
Like this 
```
    "WriteTo": [
      {
        "Name": "Elasticsearch",
        "Args": {
          "nodeUris": "http://localhost:9200/",
          "typeName": null,
          "formatStackTraceAsArray": true
        }
      }
    ]
```
So I add paramater to `LoggerConfiguration` method

No breaking
I add paramater to the end

**Other information**:


mivano:
Thanks, although added to the back and being optional, does however still makes it breaking. It is not a binary-compatible change as the call site changes. I m a bit hesitant to accept as the method signature is already long, too long IMHO. Is there a way for you to work around it?

AiSY-Yang:
It looks like you have a problem with test script
The test is pass but have error in test report stage

The Serilog only support inject paramater in method parameters,can't modify property after constructor.
so when you want to set all properties in a configuration file , must hav a method include all properties of `ElasticsearchSinkOptions` . It can be an extension method or a constructor ,Add a constructor of `ElasticsearchSinkOptions` is same as this PR.

I just want to set everything in the configuration file,If that's hard I can continue to set parameters through the program

Issue #518, opened 2023-03-08
v9.0.0 fails template creation with ES v7 with ElasticsearchClientException


AlexGoris-KasparSolutions:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [x] Bug
- [ ] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
9.0.0

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [ ] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x
- [x] 6.0

**Please describe the current behavior?**
When starting with an ES v7 server which doesn't have any serilog template configure, the sink tries to create the template, this fails with the following exception:
```
2023-03-08T10:12:49.7376289Z Failed to create the template. Elasticsearch.Net.ElasticsearchClientException: Request failed to execute. Call: Status code 400 from: PUT /_template/serilog-events-template?include_type_name=true. ServerError: Type: illegal_argument_exception Reason: "Malformed [mappings] section for type [dynamic_templates], should include an inner object describing the mapping"
   at Elasticsearch.Net.Transport`1.HandleElasticsearchClientException(RequestData data, Exception clientException, IElasticsearchResponse response)
   at Elasticsearch.Net.Transport`1.FinalizeResponse[TResponse](RequestData requestData, IRequestPipeline pipeline, List`1 seenExceptions, TResponse response)
   at Elasticsearch.Net.Transport`1.Request[TResponse](HttpMethod method, String path, PostData data, IRequestParameters requestParameters)
   at Elasticsearch.Net.ElasticLowLevelClient.DoRequest[TResponse](HttpMethod method, String path, PostData data, IRequestParameters requestParameters)
   at Elasticsearch.Net.NamespacedClientProxy.DoRequest[TResponse](HttpMethod post, String url, PostData body, IRequestParameters params)
   at Elasticsearch.Net.Specification.IndicesApi.LowLevelIndicesNamespace.PutTemplateForAll[TResponse](String name, PostData body, PutIndexTemplateRequestParameters requestParameters)
[...]

PR #519, opened 2023-03-08
fix: Don't set include_type_name query param for ES v7 template creation call

AlexGoris-KasparSolutions:
#518 

No

**Please check if the PR fulfils these requirements**
Sorry, there are no unit tests around the changed code, and making them would require me to mock the `ElasticLowLevelClient` that's being used there, which would require more time then I can spend on this.

**Other information**:
The previous code (treat all <8 version the same) would have been fine if the `TypeName` would have had a valid value, but I noticed that in the constructor of the `ElasticsearchSinkState` class we have some code which sets  its value to `null` for anything higher or equal to ES v7:
```
// Resolve typeName
if (_versionManager.EffectiveVersion.Major < 7)
    _options.TypeName = string.IsNullOrWhiteSpace(_options.TypeName)
        ? ElasticsearchSinkOptions.DefaultTypeName // "logevent"
        : _options.TypeName;
else
    _options.TypeName = null;
```
Therefor you had a scenario for ES v7 where the `TypeName` would be `null`, but the `include_type_name` query parameter would be set to `true`, which wasn't a valid API call for ES v7

### DIFF
--- a/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkState.cs
+++ b/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkState.cs
@@ -172,20 +172,17 @@ namespace Serilog.Sinks.Elasticsearch
                     }
                 }
 
-                StringResponse result;                
-                if (_versionManager.EffectiveVersion.Major < 8)
-                { 
-                    result = _client.Indices.PutTemplateForAll<StringResponse>(_templateName, GetTemplatePostData(),
+                var result = _versionManager.EffectiveVersion.Major switch
+                {
+                    < 7 => _client.Indices.PutTemplateForAll<StringResponse>(_templateName, GetTemplatePostData(),
                         new PutIndexTemplateRequestParameters
                         {
                             IncludeTypeName = IncludeTypeName ? true : (bool?)null
-                        });
-                }
-                else
-                {
+                        }),
+                    < 8 => _client.Indices.PutTemplateForAll<StringResponse>(_templateName, GetTemplatePostData()),
                     // Default to version 8 API
-                    result = _client.Indices.PutTemplateV2ForAll<StringResponse>(_templateName, GetTemplatePostData());
-                }
+                    _ => _client.Indices.PutTemplateV2ForAll<StringResponse>(_templateName, GetTemplatePostData()),
+                };
 
                 if (!result.Success)
                 {


nenadvicentic (changes requested):
This line is the the actual issue, because condition somehow got reversed from `<` to `>=` during refactoring for final `9.0.0`:

https://github.com/serilog-contrib/serilog-sinks-elasticsearch/blob/a01baffec9a74a06ab7d1beb5791e98da9989c3e/src/Serilog.Sinks.Elasticsearch/Sinks/ElasticSearch/ElasticsearchSinkState.cs#L55

Could be fixed something like this (+ removing private member above completely):
```csharp
StringResponse result;                
if (_versionManager.EffectiveVersion.Major < 8)
{ 
    result = _client.Indices.PutTemplateForAll<StringResponse>(_templateName, GetTemplatePostData(),
        new PutIndexTemplateRequestParameters
        {
            IncludeTypeName = _versionManager.EffectiveVersion.Major < 7 ? true : null
        });
}
else
{
    // Default to version 8 API
    result = _client.Indices.PutTemplateV2ForAll<StringResponse>(_templateName, GetTemplatePostData());
}
```
Or, if you prefer `switch` expression, please make it a bit simpler, so it's easier to read:

```csharp
var result = _versionManager.EffectiveVersion.Major switch
{
    < 8 => _client.Indices.PutTemplateForAll<StringResponse>(_templateName, GetTemplatePostData(),
        new PutIndexTemplateRequestParameters
        {
            IncludeTypeName = _versionManager.EffectiveVersion.Major < 7 ? true : null
        }),
    // Default to version 8 API
    _ => _client.Indices.PutTemplateV2ForAll<StringResponse>(_templateName, GetTemplatePostData()),
};
```

It would also be great if you can add this check to unit tests after check for Path:

https://github.com/serilog-contrib/serilog-sinks-elasticsearch/blob/a01baffec9a74a06ab7d1beb5791e98da9989c3e/test/Serilog.Sinks.Elasticsearch.Tests/Templating/Sendsv6TemplateTests.cs#L45

And in v7 and v8 to ensure that `uri.Query` should not contain `include_type_name` parameter.

RefaelP23:
Any updates on this PR?
The regression introduced in the `IncludeTypeName` is breaking some of our logs to Elastic.

nenadvicentic:
@RefaelP23 Yes, this was coding error. @AlexGoris-KasparSolutions did not yet reply to my review suggestions.
@mivano I believe this fix should be `9.0.1` release, before we switching to `v10`.

RefaelP23:
@nenadvicentic Thanks for the prompt reply.

I also saw your discussion here [Added original exception to FailureCallback action](https://github.com/serilog-contrib/serilog-sinks-elasticsearch/commit/a01baffec9a74a06ab7d1beb5791e98da9989c3e) and I understand that it might take some time until it's released.
Is there some workaround I can try until it's released?

nenadvicentic:
@RefaelP23 From the top of my head, I can only think of this two options:

1. Set configuration option `AutoRegisterTemplate` to `false`. I think this should suffice. I think Elasticsearch v7 is working just fine without using custom template.
2. Manually create index template on the Elasticsearch servers/clusters where you have the issue, using PostMan, curl, or similar tool. Perhaps to take exact template JSON from the servers where you currently don't have issue and put it to the once where you have the issue. You can list all templates via `/_cat/templates` and find the one related with this sink.

mivano:
@nenadvicentic this one can be removed/closed now?

nenadvicentic:
@mivano Yes. This can be deleted. I implemented in #533  what I described in this code review

Issue #384, opened 2021-03-04
Elastricsearch with self signed certificates SSL


YassineHk:
**A few questions before you begin:**

> Is this an issue related to the [Serilog core project](https://github.com/serilog/serilog) or one of the [sinks](https://github.com/serilog/serilog/wiki/Provided-Sinks) or 
[community projects](https://github.com/serilog/serilog/wiki/Community-Projects).  
This issue list is intended for Serilog Elasticsearch Sink issues. If this issue relates to another sink or to the code project, 
please log on the related repository.  Please use [Gitter chat](https://gitter.im/serilog/serilog) and [Stack Overflow](http://stackoverflow.com/questions/tagged/serilog)  for discussions and questions.

**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [X] New Feature

**What version of Serilog.Sinks.Elasticsearch is affected?  Please list the related NuGet package.**
latest

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [X] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**the current behavior**
I want to pass an SSL certificate generated by myself from my dotnet app to my certificate elasticsearch

**the expected behavior**
`
var elasticUri = Configuration["ElasticConfiguration:Uri"];
StreamWriter writer = File.CreateText("serilog.log");
SelfLog.Enable(writer); 
Log.Logger = new LoggerConfiguration() .Enrich.FromLogContext()
.MinimumLevel
.Information() 
.WriteTo.RollingFile("Logs/log-{Date}.log", LogEventLevel.Warning)
.WriteTo.Elasticsearch(new ElasticsearchSinkOptions(new Uri(elasticUri)) 
{
    IndexFormat = Configuration["ElasticConfiguration:Index"],
    AutoRegisterTemplate = true,
})
.CreateLogger(); 
`

tobiasschittkowski:
You can use the following option:

                ModifyConnectionSettings = c => c
                    .ConnectionLimit(-1)
                    .ServerCertificateValidationCallback((o, certificate, arg3, arg4) => { return true; })

That will override the server certificate validation.

PR #562, opened 2023-07-25
Add Sink configure option to effectively disable SSL Certificate Verification

jasonsummers:
Add the 'ignoreServerCertificateValidation' Sink configuration parameter to LoggerConfigurationElasticSearchExtensions.cs which, when true will configure the Elasticsearch client with a ServerCertificateValidationCallback which always returns true.

The default behaviour is unchanged.

#384 

No

- [ ] Unit Tests for the changes have been added (for bug fixes / features) - Could not find a sensible place to put tests, grateful for direction if tests are required for this change.

**Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,8 @@ In your `appsettings.json` file, under the `Serilog` node, :
           "connectionPool": "My.Namespace.MyConnectionPool, My.Assembly.Name",
           "customFormatter": "My.Namespace.MyCustomFormatter, My.Assembly.Name",
           "customDurableFormatter": "My.Namespace.MyCustomDurableFormatter, My.Assembly.Name",
-          "failureSink": "My.Namespace.MyFailureSink, My.Assembly.Name"
+          "failureSink": "My.Namespace.MyFailureSink, My.Assembly.Name",
+          "ignoreServerCertificateValidation": "false"
         }
     }]
   }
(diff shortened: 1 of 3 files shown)


mivano:
Thanks for the PR. As it is an extra parameter (even with a default value), it still is a breaking change. Not a big issue though. More troublesome; this introduces a risk as you should check the certificates and not just return `true`. You most likely have a good reason for this change. Is there another way to get the behaviour as I m not that in favour of adding this option to the already large list of parameters?

jasonsummers:
Hi @mivano,

Thank you very much for the prompt and useful feedback.
I've pushed another commit which addresses the concern of the growing parameter list by combining the new SSL options with the existing ConnectionGlobalHeaders option. This approach is clearly more of a breaking change but does keep the number of parameters the same.
I've also made the settings more granular to ignore specific types of SSL Error.

I'd appreciate your guidance on what you feel would be an acceptable solution from a security perspective, I've come up with the following options and would appreciate your perspective:

1. Add an 'AllowedSignatures' option to the new configuration where consumers can place something like the hash or the public string of the certificates that they would like to ignore.
2. Scan the current working directory for certificates which should be allowed (arguably removes the need for a change in the configuration options altogether)

How deep do you think we should go from a security perspective?
The security risk as far as I can see is that  a threat actor could redirect the log traffic (and potentially sensitive information) to their own elastic search instance. In order for such an attack to be successful the attacker would need to either:
a) compromise appsettings.json and change the nodeUris property to point to their Elastic search instance; or
b) compromise the DNS/IP of the current Elastic search instance which would only be successful if the target had one or more of the new 'IgnoreSsl' options set to true.

If either of these scenarios occurred I'd possibly argue that the target has bigger problems. Very grateful for your thoughts on this.

mivano:
I like this new approach. Makes it more concise what needs to be enabled and set up. For me, the simple switch of setting some boolean was what I not really liked, this solution requires more explicit steps. And you are right; it is never completely securable. It is however more of a breaking change, so I will need to document that.

